PR #25, opened 2022-01-10
Figure for an ARG with no recomb nodes

### LINKED ISSUE #17: Recombination nodes are optional figure
We have  placeholder figure (``fig-recombination-nodes``) whose caption current reads:

> Recombination nodes are optional in an edge-annotated ARG. (A) A simulated ARG including all recombination nodes. (B) The same ARG with recombination nodes removed. (C) The corresponding sequence of trees.

I'm imaging that do a small simulation with msprime's ``record_full_arg`` (with the single recombination node encoding) and we draw the graph in (A). We then simplify this tree sequence to remove the recombination nodes, keeping the node IDs constant for clarity, and draw this in (B). We then use ``draw_svg`` to draw the tree sequence in the standard form and show this in (C).

Since this is (probably?) two long things and one wide thing it may be best to do this as a two-row display like this?
```
 ---------
| A | B |
|   C    |
---------
```

@hyanwong, this seems right up your street - can I assign it to you?

hyanwong:
Fixes #17


paper.tex
```diff
@@ -190,11 +190,24 @@ \subsection*{Edge annotated recombination}
...
+ARG including all recombination nodes (red). (b) The same ARG having removed both
+recombination nodes and nodes of common ancestors in which no genetic
+lineages coalesce (cyan). (c) The sequence of trees corresponding to (b): if we were
+to plot out the trees corresponding to (a) we would have more trees, but the topologies
+and MRCA nodes at each genomic location would be identical. For clarity, nodes are
+plotted by ranked time rather than true time on the y-axis.
+% HYW: note that we also truncate nodes for portions in which they are unary,
+% so that nodes which are coalescent nodes over only some of their length
+% (blue-green) become coalescent nodes over their entire length (green). This
+% can, however, throw away some useful information, for example node 4 has
```
jeromekelleher, line 207: This is what we do in tsinfer essentially isn't it? So perhaps worth noting somewhere afterall?

jeromekelleher:
I would really prefer not to put the PDFs in if we don't need to. Basically every parallel change to the repo requires a merge conflict resolution once you do, so unless there's a compelling reason for keeping them in at this point, could we remove them please? (I agree it makes sense to put them in later in the process)

paper.tex
jeromekelleher: As in, when we simplify with ``keep_unary=True``

hyanwong:
> I would really prefer not to put the PDFs in if we don't need to

Oh, sorry, I see. Yes, of course: no reason to include them. I didn't notice that we had a makefile, so I'll rework this PR to use that too.

hyanwong:
> > I would really prefer not to put the PDFs in if we don't need to
> 
> Oh, sorry, I see. Yes, of course: no reason to include them. I didn't notice that we had a makefile, so I'll rework this PR to use that too.

Oh, I think I misunderstood. You mean the paper pdf right? Do you also want to exclude the figure/illustration pdfs & svgs @jeromekelleher ?

paper.tex
hyanwong: Yes, this is what we do. The discussion is at https://github.com/tskit-dev/tskit/discussions/2089

jeromekelleher:
> Oh, I think I misunderstood. You mean the paper pdf right? Do you also want to exclude the figure/illustration pdfs & svgs

Any binaries are a pain, so the figure PDFs and the main text pdf.

Suppose I run "make" and it decides to rebuild all the illustration PDFs (for whatever reason). Then, all the binaries will show up as changed, even though they have no meaningful differences (pdfs must encode a timestamp or something). Then, I either have to be careful to drop them from the commit, or just commit the meaningless differences and (a) bloat the repo and (b) have merge conflicts if someone else happens to have done the same thing. 

No binaries, please, unless it's really genuinely necessary. Git just isn't designed for it, and it makes a lot of annoying pointless work for maintainers.

jeromekelleher:
The svgs are probably handy to store on the other hand, that will save the requirement on Python 3.10 (which I think you have?)

jeromekelleher:
I think it's probably better to commit the SVGs all right, it's a bit of a faff getting the dependencies setup for pygraphviz and graphlib, and this will make it easier for people to make the document.

PR #27, opened 2022-01-11
Initial requirements.txt

jeromekelleher:
This made the figures build for me - I guess we should add in networkx and a few more things @hyanwong?


hyanwong:
Yep, shall we commit this and I'll add some more?

PR #39, opened 2022-01-21
Arg sim

jeromekelleher:
I think we've made some real progress here! There's various things to talk about, but for now, here's the high-level stuff.

This PR adds some functions to simulate an ARG in two different encodings, and a function to convert one encoding into the other. The first encoding is what I'm calling "ancestry-resolved", and returns a well formed tskit TreeSequence (complete with all ARG nodes). The second encoding is non-resolved, close to what we'd think of as a classical Griffiths ARG: it just stores the nodes, (graph) edges between them and recombination breakpoints. The third function then converts the unresolved ARG into a full resolved one, so that we can get precisely the same final tree sequence by running the two different simulations under the same parameters. There's a bijection between the two forms.

A key insight here is that simulating an unresolved ARG is fundamentally pointless, since you *must* already have all the information required to output a resolved ARG in order to run the simulation. Unless you want to simulate back to the GMRCA (and filter out many pass-through nodes), you must keep track of the amount of ancestral material remaining in intervals along the genome, so that they can be "snipped out" once fully coalesced. It is precisely this information that we need in order to output an ancestry-resolved ARG. The process of resolving this Griffiths-like representation is then simply a case of "rerunning" the simulation, keeping track of the amount of ancestral material present along the genome so that we can record tree sequence edges.

There's some more notes in the code, which is hopefully (mostly) reasonably readable.

One really important thing that I realised here is that we *must* have two nodes for a recombination - otherwise, there is no bijection. You can't convert to a fully resolved ARG from an unresolved ARG with a single recombination node because there are (I think?) 2^k different resolutions consistent with it, corresponding to the two different ways that ancestry can go at each recombination breakpoint. We usually imagine that we adopt some rule to work around this, but I couldn't figure out any way to do it that would work for simulations. Anyway, a single recombination node is fundamentally unbiological when you think it through - I'll open a discussion to talk about this more.

Good to get your eyes on this @hyanwong and @JereKoskela.

### DIFF
--- a/utils/args.py
+++ b/utils/args.py
@@ -167,20 +167,25 @@ def arg_sim(n, rho, L, seed=None):
         t_inc = min(t_re, t_ca)
         t += t_inc
         if t_inc == t_re:
-            lineage = rng.choices(lineages, weights=lineage_links)[0]
-            breakpoint = rng.randrange(lineage.left + 1, lineage.right)
-            assert lineage.left < breakpoint < lineage.right
-            node = tables.nodes.add_row(
+            left_lineage = rng.choices(lineages, weights=lineage_links)[0]
+            breakpoint = rng.randrange(left_lineage.left + 1, left_lineage.right)
+            assert left_lineage.left < breakpoint < left_lineage.right
+            child = left_lineage.node
+            left_parent = tables.nodes.add_row(
                 flags=NODE_IS_RECOMB, time=t, metadata={"breakpoint": breakpoint}
             )
-            right = lineage.split(breakpoint)
-            lineages.append(right)
-            for lineage in [lineage, right]:
+            right_parent = tables.nodes.add_row(
+                flags=NODE_IS_RECOMB, time=t, metadata={"breakpoint": breakpoint}
+            )
+            right_lineage = left_lineage.split(breakpoint)
+            left_lineage.node = left_parent
+            right_lineage.node = right_parent
+            for lineage in [left_lineage, right_lineage]:
                 for interval in lineage.ancestry:
                     tables.edges.add_row(
-                        interval.left, interval.right, node, lineage.node
+                        interval.left, interval.right, lineage.node, child
                     )
-                lineage.node = node
+            lineages.append(right_lineage)
         else:
             a = lineages.pop(rng.randrange(len(lineages)))
             b = lineages.pop(rng.randrange(len(lineages)))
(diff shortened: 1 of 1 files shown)


hyanwong:
> One really important thing that I realised here is that we _must_ have two nodes for a recombination - otherwise, there is no bijection. You can't convert to a fully resolved ARG from an unresolved ARG with a single recombination node because there are (I think?) 2^k different resolutions consistent with it, corresponding to the two different ways that ancestry can go at each recombination breakpoint. We usually imagine that we adopt some rule to work around this, but I couldn't figure out any way to do it that would work for simulations.

I might have misunderstood here (and I've only skimmed through the code so far), but the "normal" way to decide which way the ancestry goes is to use the order of the edges. The first edge from the RE node takes the material to the left of the breakpoint, the second takes the material to the right.

If you use 2 nodes, then (in practice) you rely on the order of the nodes instead. The first node is the stuff to the left, the second to the right. If you have `edge.left` and `edge.right` on these nodes then in a simple Griffiths graph you could use those to work out which takes which material, but in the unresolved ARG you don't have that information anyway.

I worry (quite a lot!) that the 2 node version is simply an easy way to encode a Griffiths graph, and that it isn't actually any more useful than the single node version, once we start having multiple recombination breakpoints. In fact, it is potentially problematic as it encourages the user to think of the "wrong way" to solve the problem of identifying the ancestral paths.

JereKoskela:
I think I follow what's going on, but could well be wrong. For a simple example, lets take 2 samples and 2 sites. Suppose the first two events are a recombination at each sample, splitting the all four ancestral sites into distinct lineages. Then suppose that one pair of sites merges to one local MRCA, and the other to another.

From the resulting Griffiths ARG, there is no way to tell which MRCA belongs to which site. As @hyanwong says, you have to rely on some sort of labelling. I don't think we rely on edge ordering as such; rather, the parent of the edge with the leftmost `edge.left` tells us which way to resolve the recombination. But that information isn't available in the Griffiths ARG, breaking the bijection. Is that right?

What I don't see is why two recombination nodes helps - I think you'd still have the same issue.

I also don't think one recombination node is any less biological than two. The difference is whether we store the last generation before the recombination (backwards in time), or the first one after.

jeromekelleher:
I'll think this through more. The issue that it seemed like I was hitting up against was that an unambigous (in terms of the flow of ancestry) Griffiths graph could not be simulated. I think there are some fundamental identifiability issues with the Griffiths graph, and maybe this is a good angle to look at that from.

hyanwong:
> From the resulting Griffiths ARG, there is no way to tell which MRCA belongs to which site. As @hyanwong says, you have to rely on some sort of labelling. I don't think we rely on edge ordering as such; rather, the parent of the edge with the leftmost `edge.left` tells us which way to resolve the recombination. 

That's what we do when we encode the information on the edges.

> But that information isn't available in the Griffiths ARG, breaking the bijection. Is that right?

In the true Griffiths ARG, with node annotations but no edge annotations, since we aren't storing `edge.left`, we must rely on something else. I think the normal thing is to use the defined order of the edges

jeromekelleher:
I'm going to merge this much - I'm convinced the machinery here is useful, and we'll use it somewhere.

PR #42, opened 2022-01-25
Initial theory + Monte Carlo writeup

JereKoskela:
Some rather basic connective tissue between some references on ARGs from a theoretical perspective, and on some inference methods for ARGs. This is undoubtedly biased towards material I happen to have read, so getting some more diverse eyes on it would be good. In particular, the paragraph on heuristic and parsimony-based methods is basically just a citation blob.

I'm probably not using terminology consistently with what we will eventually want to emphasize, but figured it was best to get this up so others (especially @a-ignatieva) can take over, rather than obsess with it myself for longer.


paper.tex
```diff
@@ -68,28 +68,62 @@ \section*{Ancestral graphs and stochastic processes}
...
+
+In a recombining organism, the genetic ancestry of a longer segment of DNA is 
+no longer described by a single tree. Instead, ancestral relationships in a
+sample can be represented by a sequence of correlated, marginal trees along the
+genome. The Ancestral Recombination Graph (ARG) was introduced by
+Griffiths~\citep{griffiths1991two,griffiths1997ancestral} as a branching-coalescing 
+stochastic process whose realisations are graphs which into which all of the
+marginal ancestral trees are embedded. Coalescences encode common ancestry
+as they do in the coalescent, while branching events encode points at which a
+marginal tree changes from one position along the genome to the next.
```
jeromekelleher, line 85:
```suggestion
marginal tree changes from one position along the genome to the next.
%note-jk: could nitpick this in that diamonds don't generate tree "changes" as such. Maybe tweak this on the next round of revision?
```

paper.tex
```diff
@@ -68,28 +68,62 @@ \section*{Ancestral graphs and stochastic processes}
...
+
+There are several formulations of the ARG, which vary by
```
jeromekelleher, line 87:
```suggestion
% jk-note: this isn't quite right I think, as it's muddling the distinction between a graph stochastic process 
% definition with the graph as the encoding of the marginal trees. 
There are several formulations of the ARG, which vary by
```
I don't think this is quite right, but I can't see how to tweak it immediately. I suggest we put in the comment and look at it again when we have more text to connect with.

paper.tex
```diff
@@ -68,28 +68,62 @@ \section*{Ancestral graphs and stochastic processes}
...
+the amount of extra information they capture beyond a minimal representation
+of all marginal trees. Two prominent examples are the ``big''
+ARG~\citep{ethier1990two}, which has a particularly simple specification as a
+stochastic process but is computationally infeasible, and the ``little ARG''
+traversed by Hudson's algorithm~\citep{hudson1983properties}, which
```
jeromekelleher, line 92: Didn't Griffiths work with the little ARG too?

paper.tex
```diff
@@ -217,20 +251,46 @@ \section*{ARG inference methods}
...
-\item Discussions of the complexity of inferring ARGs (but we need to
-dig into this: is this the complexity of sampling from the coalescent,
-finding a ``min ARG'', or something else?
-\citep{bordewich2005computational,wang2001perfect,hein1990reconstructing,song2005constructing}
-\item Recent methods: Relate, tsinfer, tsdate, etc.
-\end{itemize}
+The ARG has been used as a model for recombining DNA sequence data essentially
+since it was first formulated. Early methods focused on finding parsimonious
+ARGs consistent with a given set of
+sequences \citep{hein1990reconstructing, lyngso2005minimum}.
```
jeromekelleher, line 257: I don't think hein1990 reffered to ARGs at all.

paper.tex
```diff
@@ -217,20 +251,46 @@ \section*{ARG inference methods}
...
+and the underlying data structure, but the most efficient methods can be practically
+used on megabase scale data and tens of thousands of samples under human-like
+parameters \citep{kelleher2019inferring, zhang2021biobank}.
+
+An alternative approach is to treat the ancestry as a latent parameter to be averaged out
+by Monte Carlo methods, based either on importance sampling 
+\citep{griffiths1996ancestral, fearnhead2001estimating, jenkins2011inference} 
+or MCMC \citep{kuhner2000maximum, nielsen2000estimation, wang2008bayesian}.
+These methods invariably operated on a representation of the ``little ARG", typically
+generated until the grand MRCA. They are extremely computationally expensive,
```
jeromekelleher, line 273: Really? As in, the first common ancestor of all nodes in the Big ARG?

jeromekelleher (commented):
This is great, thanks @JereKoskela!

The narrative will need some tweaking as we get other parts of the paper up to speed, but this is a really excellent starting point.

I've made a few comments along the way, most are "suggestions" which should add the latex comment into the text (if you accept them).

paper.tex
hyanwong: I agree. Good spot @jeromekelleher

hyanwong:
This is *great*, thanks Jere.

paper.tex
```diff
@@ -68,28 +68,62 @@ \section*{Ancestral graphs and stochastic processes}
...
+lineages~\citep{etheridge2009coalescent} or by leveraging perfect simulation
+techniques when recurrent mutation is present\citep{fearnhead2001perfect}.
+Extensions of the ASG have been developed to frequency-dependent 
+selection~\citep{neuhauser1999ancestral, gonzalezcasanova2018duality},
+unlinked chromosomes~\citep{fearnhead2003ancestral}, recombining
+loci whereupon branching is due to both selection and
+recombination~\citep{donnelly1999genealogical}, and high fecundity 
+reproduction~\citep{gonzalezcasanova2018duality, koskela2019robust}.
 
 In the theoretical literature, ancestral graphs such as the ARG refer to the
```
hyanwong, line 125: This is not a sentence modified in this PR, but I think we agreed that in (most of) the theoretical literature, the ARG is assumed to mean *both* the process and the structure generated by it.

paper.tex
JereKoskela: Just changing the verb from "capture" to "generate", or something to that effect. Either way, I agree this needs clarifying.

paper.tex
JereKoskela: I'm beginning to think I'm not 100% clear on the details of what "big" and "little" ARGs mean. I think we should definitely have a quick summary.

paper.tex
JereKoskela: Ah, no. I might be misusing GMRCA. I mean the first node which is a common ancestor at all sites, without pruning away sites as soon as their local MRCA is reached (see Figure 1 [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1461855/pdf/11729171.pdf)). I think I should push one more commit with a slightly longer description to that effect, and some partial tweaks based on the comments above.

jeromekelleher: That's my understanding of the GMRCA - there's nothing grand about the last node that happens to be an MRCA in the little ARG.

paper.tex
jeromekelleher: I agree - this is something we'll have to tease out more carefully as it's all very confusing. We don't need to do it here though, let's get this much down on paper first.

hyanwong: I agree - let's put in a comment to that effect in the latex and then merge this.

PR #43, opened 2022-01-27
Wright-Fisher ARG simulator

jeromekelleher:
Here's a very rough simulator for ARGs in the Wright Fisher process. It was quite the mindy-bendy task, but well worth it I think.

The main thing I've gotten from this is a sense that the "standard" ARG (unsurprisingly) derives very closely from the CwR process, and it's particular structure owes a lot to the assumptions made in that process. The neat classification of nodes that you get from the coalescent really  doesn't hold any more when you have a small population WF process.

Basically you get everything happening at once in all possible combinations in the simulation. 

```
27.00┊   87       ┊            ┊            ┊            ┊                                                                                                                          
     ┊ ┏━━┻━━┓    ┊            ┊            ┊            ┊
26.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
25.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
24.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
22.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
21.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
20.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
19.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
17.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
16.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
15.00┊ ┃     ┃    ┊            ┊            ┊   66       ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┏━━┻━━┓    ┊
14.00┊ R64   ┃    ┊            ┊            ┊ ┃    R63   ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
13.00┊ 61   R60   ┊            ┊            ┊ R59  61    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
12.00┊ ┃    55    ┊            ┊            ┊ 55    ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
11.00┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
10.00┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
9.00 ┊ ┃    R46   ┊            ┊            ┊ R45   ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
8.00 ┊ ┃    43    ┊            ┊            ┊ 43    ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
7.00 ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
6.00 ┊ ┃     ┃    ┊    33      ┊    33      ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊ ┏━━━╋━━━┓  ┊ ┏━━━╋━━━┓  ┊ ┃     ┃    ┊
5.00 ┊ ┃     ┃    ┊ R29R26  ┃  ┊ R29R26  ┃  ┊ R30  R27   ┊
     ┊ ┃     ┃    ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
4.00 ┊ R20  R23   ┊ R22 ┃  R24 ┊ R21R19 R24 ┊ R21  R19   ┊
     ┊ ┃    ┏┻━┓  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
3.00 ┊ R14 R18 ┃  ┊ R15R17  ┃  ┊ R15R14  ┃  ┊ R15  R14   ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃    ┏┻━┓  ┊
2.00 ┊ R8  R13R11 ┊ R8 R13 R11 ┊ R8 R12 R11 ┊ R9  R10R12 ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃    ┃  ┃  ┊
1.00 ┊ R5  R7 R6  ┊ R4 R7  R6  ┊ R4 R7  R6  ┊ R4  R6 R7  ┊
     ┊ ┃   ┏┻┓ ┃  ┊ ┃   ┃  ┏┻┓ ┊ ┃   ┃  ┏┻┓ ┊ ┃   ┏┻┓ ┃  ┊
0.00 ┊ 0   1 3 2  ┊ 0   1  2 3 ┊ 0   1  2 3 ┊ 0   1 3 2  ┊
     0            1            2            3            4

```

### DIFF
--- a/utils/args.py
+++ b/utils/args.py
@@ -193,8 +193,14 @@ def merge_ancestry(lineages):
         interval = AncestryInterval(left, right, ancestral_to)
         yield interval, [u.value[0] for u in U]
 
+@dataclasses.dataclass
+class Node:
+    time: float
+    flags: int = 0
+    metadata: dict = dataclasses.field(default_factory=dict)
 
-def arg_sim(n, rho, L, seed=None):
+
+def arg_sim(n, rho, L, seed=None, include_parent_nodes=False):
     """
     Simulate an ancestry-resolved ARG under the coalescent with recombination
     and return the tskit TreeSequence object.
@@ -205,14 +211,13 @@ def arg_sim(n, rho, L, seed=None):
     tables = tskit.TableCollection(L)
     tables.nodes.metadata_schema = tskit.MetadataSchema.permissive_json()
     lineages = []
+    nodes = []
     for _ in range(n):
-        node = tables.nodes.add_row(
-            time=0,
-            flags=tskit.NODE_IS_SAMPLE,
-        )
-        lineages.append(Lineage(node, [AncestryInterval(0, L, 1)]))
+        lineages.append(Lineage(len(nodes), [AncestryInterval(0, L, 1)]))
+        nodes.append(Node(time=0, flags=tskit.NODE_IS_SAMPLE))
 
     t = 0
+
     while len(lineages) > 0:
         # print(f"t = {t:.2f} k = {len(lineages)}")
         # for lineage in lineages:
@@ -233,33 +238,34 @@ def arg_sim(n, rho, L, seed=None):
             assert left_lineage.left < breakpoint < left_lineage.right
             right_lineage = left_lineage.split(breakpoint)
             child = left_lineage.node
-            for lineage in left_lineage, right_lineage:
-                lineage.node = tables.nodes.add_row(
-                    flags=NODE_IS_RECOMB, time=t, metadata={"breakpoint": breakpoint}
-                )
-                for interval in lineage.ancestry:
-                    tables.edges.add_row(
-                        interval.left, interval.right, lineage.node, child
-                    )
+            nodes[child].flags |= NODE_IS_RECOMB
+            nodes[child].metadata["breakpoint"] = breakpoint
+            if include_parent_nodes:
+                for lineage in left_lineage, right_lineage:
+                    lineage.node = len(nodes)
+                    nodes.append(Node(time=t))
+                    for interval in lineage.ancestry:
+                        tables.edges.add_row(
+                            interval.left, interval.right, lineage.node, child
+                        )
             lineages.append(right_lineage)
         else:
             a = lineages.pop(rng.randrange(len(lineages)))
             b = lineages.pop(rng.randrange(len(lineages)))
-            c = Lineage(len(tables.nodes), [])
-            flags = NODE_IS_NONCOAL_CA
+            c = Lineage(len(nodes), [])
             for interval, intersecting_lineages in merge_ancestry([a, b]):
-                if len(intersecting_lineages) > 1:
-                    flags = 0  # This is a coalescence, treat this as ordinary tree node
                 if interval.ancestral_to < n:
                     c.ancestry.append(interval)
                 for lineage in intersecting_lineages:
                     tables.edges.add_row(
                         interval.left, interval.right, c.node, lineage.node
                     )
-            tables.nodes.add_row(flags=flags, time=t, metadata={})
+            nodes.append(Node(time=t))
             if len(c.ancestry) > 0:
                 lineages.append(c)
 
+    for node in nodes:
+        tables.nodes.add_row(flags=node.flags, time=node.time, metadata=node.metadata)
     tables.sort()
     return tables.tree_sequence()
 
@@ -351,12 +357,9 @@ class Individual:
     )
 
 
-def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
+def resolved_wf_arg_sim(n, N, L, seed=None):
     """
     NOTE! This hasn't been statistically tested and is probably not correct.
-
-    If gametes=True, store RE nodes for the recombinant gametes rather
-    than the individual genomes.
 
     We don't keep track of the pedigree because this is inconsistent
     with the practise of dropping "pass through" nodes in which
@@ -585,9 +588,11 @@ def simplest_example():
 
 # simplest_example()
 
-ts = resolved_wf_arg_sim(2, 2, 4, 46, gametes=False)
-print(ts.tables)
-draw_arg(ts)
+# ts = resolved_wf_arg_sim(2, 2, 4, 46)
+for include_parent_nodes in [False, True]:
+    ts = arg_sim(6, 0.1, 5, 46, include_parent_nodes=include_parent_nodes)
+# print(ts.tables)
+    draw_arg(ts)
 
 
 # n = 2
--- a/utils/args.py
+++ b/utils/args.py
@@ -20,10 +20,16 @@ def draw_arg(ts):
     node_labels = {}
     for node in ts.nodes():
         label = str(node.id)
-        if node.flags == NODE_IS_RECOMB:
+        if node.flags & NODE_IS_RECOMB > 0:
             label = f"R{node.id}"
         elif node.flags == NODE_IS_NONCOAL_CA:
             label = f"N{node.id}"
+        if "total_span" in node.metadata:
+            total_span = node.metadata["total_span"]
+            coal_span = node.metadata["coal_span"]
+            # Putting in an extra space at the end as a quick hack
+            # to workaround spacing problems.
+            label += f":{coal_span}/{total_span} "
         node_labels[node.id] = label
     print(ts.draw_text(node_labels=node_labels))
 
@@ -50,6 +56,10 @@ class AncestryInterval:
     left: int
     right: int
     ancestral_to: int
+
+    @property
+    def span(self):
+        return self.right - self.left
 
 
 @dataclasses.dataclass
@@ -196,7 +206,10 @@ def arg_sim(n, rho, L, seed=None):
     tables.nodes.metadata_schema = tskit.MetadataSchema.permissive_json()
     lineages = []
     for _ in range(n):
-        node = tables.nodes.add_row(time=0, flags=tskit.NODE_IS_SAMPLE)
+        node = tables.nodes.add_row(
+            time=0,
+            flags=tskit.NODE_IS_SAMPLE,
+        )
         lineages.append(Lineage(node, [AncestryInterval(0, L, 1)]))
 
     t = 0
@@ -354,14 +367,18 @@ def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
     tables.nodes.metadata_schema = tskit.MetadataSchema.permissive_json()
 
     ancestors = []
+    node_flags = []
     for _ in range(n):
         ind = Individual(tables.individuals.add_row(), [])
         for _ in range(2):
+            # Arbitrarily setting coal span to L
             node = tables.nodes.add_row(
-                time=0, flags=tskit.NODE_IS_SAMPLE, individual=ind.id
+                time=0, individual=ind.id, metadata={"coal_span": L, "total_span": L}
             )
             ind.lineages.append(Lineage(node, [AncestryInterval(0, L, 1)]))
+            node_flags.append(tskit.NODE_IS_SAMPLE)
         ancestors.append(ind)
+
     t = 0
     while len(ancestors) > 0:
         t += 1
@@ -374,7 +391,6 @@ def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
         # in the population to the lineages inherited from is maternal
         # and paternal genomes
         parents = {}
-        recombinant_nodes = set()
         for ancestor in ancestors:
             for lineage in ancestor.lineages:
                 parent_index = rng.randrange(N)
@@ -391,19 +407,7 @@ def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
                     right_lineage = lineage.split(breakpoint)
                     parent.collected_lineages[j].append(right_lineage)
                     j += 1
-                    if gametes:
-                        # NOTE: This code path is exploratory, just to see what the
-                        # result looks like. Probably should be removed.
-                        # Add in a new node to represent the RE event
-                        node = tables.nodes.add_row(time=t - 0.5, flags=NODE_IS_RECOMB)
-                        for lin in lineage, right_lineage:
-                            for interval in lin.ancestry:
-                                tables.edges.add_row(interval.left, interval.right,
-                                        node, lin.node)
-                            lin.node = node
-                    else:
-                        recombinant_nodes.add(lineage.node)
-
+                    node_flags[lineage.node] |= NODE_IS_RECOMB
                 parent.collected_lineages[j].append(lineage)
 
         # All the ancestral material has been distributed to the parental
@@ -412,7 +416,6 @@ def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
 
         # print("|P| = ", len(parents), "RE_children = ", recombinant_nodes)
         for parent in parents.values():
-            # print("\t", parent)
             for lineages in parent.collected_lineages:
                 # These are all the lineages that have been collected
                 # together for this lineage on this individual. If there
@@ -422,37 +425,40 @@ def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
                 # for lin in lineages:
                 #     print("\t\t", lin)
 
-                # Debateable whether the flags mean much here, but let's call
-                # it an RE node if any effective recombination happened on
-                # any of the child lineages. We could also try to detect
-                # NODE_IS_NONCOAL_CA
-                flags = 0
-                for lineage in lineages:
-                    if lineage.node in recombinant_nodes:
-                        flags = NODE_IS_RECOMB
-                if len(lineages) == 1 and flags == 0:
+                if len(lineages) == 1 and node_flags[lineage.node] == 0:
                     merged_lineage = lineages[0]
-                else:
+                elif len(lineages) > 0:
                     if parent.id == -1:
                         parent.id = tables.individuals.add_row()
-                    node = tables.nodes.add_row(
-                        time=t, flags=flags, individual=parent.id
-                    )
-                    # print("\t\tNEW NODE", node)
+                    node = len(tables.nodes)
                     merged_lineage = Lineage(node, [])
+                    node_flags.append(0)
+                    total_span = 0
+                    coal_span = 0
                     for interval, intersecting_lineages in merge_ancestry(lineages):
+                        total_span += interval.span
+                        coal_span += interval.span * (len(intersecting_lineages) > 1)
                         if interval.ancestral_to < 2 * n:  # n is *diploid* sample size
                             merged_lineage.ancestry.append(interval)
                         for child_lineage in intersecting_lineages:
                             tables.edges.add_row(
                                 interval.left, interval.right, node, child_lineage.node
                             )
+                    tables.nodes.add_row(
+                        time=t,
+                        individual=parent.id,
+                        metadata={"total_span": total_span, "coal_span": coal_span},
+                    )
+                else:
+                    merged_lineage = Lineage(-1, [])
                 if len(merged_lineage.ancestry) > 0:
                     parent.lineages.append(merged_lineage)
             if len(parent.lineages) > 0:
                 ancestors.append(parent)
 
         assert len(ancestors) <= N
+
+    tables.nodes.flags = node_flags
 
     tables.sort()
     return tables.tree_sequence()
@@ -579,7 +585,7 @@ def simplest_example():
 
 # simplest_example()
 
-ts = resolved_wf_arg_sim(2, 5, 4, 46, gametes=False)
+ts = resolved_wf_arg_sim(2, 2, 4, 46, gametes=False)
 print(ts.tables)
 draw_arg(ts)
 
--- a/utils/args.py
+++ b/utils/args.py
@@ -338,9 +338,12 @@ class Individual:
     )
 
 
-def resolved_wf_arg_sim(n, N, L, seed=None):
+def resolved_wf_arg_sim(n, N, L, seed=None, gametes=False):
     """
     NOTE! This hasn't been statistically tested and is probably not correct.
+
+    If gametes=True, store RE nodes for the recombinant gametes rather
+    than the individual genomes.
 
     We don't keep track of the pedigree because this is inconsistent
     with the practise of dropping "pass through" nodes in which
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Here's what it looks like if we record nodes for the unique gametes (at half-generation times):
```
27.00┊   82       ┊            ┊            ┊            ┊
     ┊ ┏━━┻━━┓    ┊            ┊            ┊            ┊
26.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
25.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
24.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
22.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
21.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
20.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
19.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
17.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
16.00┊ ┃     ┃    ┊            ┊            ┊            ┊
     ┊ ┃     ┃    ┊            ┊            ┊            ┊
15.00┊ ┃     ┃    ┊            ┊            ┊   61       ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┏━━┻━━┓    ┊
13.50┊ R59   ┃    ┊            ┊            ┊ ┃    R59   ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
13.00┊ 57    ┃    ┊            ┊            ┊ ┃    57    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
12.50┊ ┃    R56   ┊            ┊            ┊ R56   ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
12.00┊ ┃    52    ┊            ┊            ┊ 52    ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
11.00┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
10.00┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
9.00 ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
8.50 ┊ ┃    R43   ┊            ┊            ┊ R43   ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
8.00 ┊ ┃    41    ┊            ┊            ┊ 41    ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
7.00 ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊            ┊            ┊ ┃     ┃    ┊
6.00 ┊ ┃     ┃    ┊    31      ┊    31      ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊ ┏━━━╋━━━┓  ┊ ┏━━━╋━━━┓  ┊ ┃     ┃    ┊
5.00 ┊ ┃     ┃    ┊ 28 26   ┃  ┊ 28 26   ┃  ┊ ┃     ┃    ┊
     ┊ ┃     ┃    ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
4.50 ┊ ┃     ┃    ┊ ┃   ┃   ┃  ┊ R25R24  ┃  ┊ R25  R24   ┊
     ┊ ┃     ┃    ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
4.00 ┊ ┃    22    ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
     ┊ ┃    ┏┻━┓  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
3.50 ┊ R19  ┃ R21 ┊ R20 ┃  R21 ┊ R20R19 R21 ┊ R20  R19   ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃     ┃    ┊
3.00 ┊ 16   ┃  ┃  ┊ 17  ┃   ┃  ┊ 17 16   ┃  ┊ 17   16    ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃    ┏┻━┓  ┊
2.50 ┊ R14 R15 ┃  ┊ R14R15  ┃  ┊ R14 ┃   ┃  ┊ ┃    ┃  ┃  ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃    ┃  ┃  ┊
2.00 ┊ 13   ┃  ┃  ┊ 13  ┃   ┃  ┊ 13  ┃   ┃  ┊ ┃    ┃  ┃  ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃    ┃  ┃  ┊
1.50 ┊ ┃   R12R11 ┊ R10R12 R11 ┊ R10R12 R11 ┊ R10 R11R12 ┊
     ┊ ┃    ┃  ┃  ┊ ┃   ┃   ┃  ┊ ┃   ┃   ┃  ┊ ┃    ┃  ┃  ┊
1.00 ┊ ┃    9  8  ┊ ┃   9   8  ┊ ┃   9   8  ┊ ┃    8  9  ┊
     ┊ ┃   ┏┻┓ ┃  ┊ ┃   ┃  ┏┻┓ ┊ ┃   ┃  ┏┻┓ ┊ ┃   ┏┻┓ ┃  ┊
0.50 ┊ R4 R5R7R6  ┊ R4 R5 R6R7 ┊ R4 R5 R6R7 ┊ R4 R5R7R6  ┊
     ┊ ┃   ┃ ┃ ┃  ┊ ┃   ┃  ┃ ┃ ┊ ┃   ┃  ┃ ┃ ┊ ┃   ┃ ┃ ┃  ┊
0.00 ┊ 0   1 3 2  ┊ 0   1  2 3 ┊ 0   1  2 3 ┊ 0   1 3 2  ┊
     0            1            2            3            4
```

jeromekelleher:
I don't really see any advantage to the gamete approach --- all it does is add some extra nodes and edges at the bottom of the trees (since we end up needing nodes for the two parental genomes anyway).

hyanwong:
The problem that I have (which led to the gamete approach) is that when there are multiple gametes per parent, which happens all the time in a WF model, then the 2 recombination nodes from one recombination are the same as the 2 recombination nodes from another recombination. So you can't store a single breakpoint on each node. You must (I think) encounter this in your WF simulator, right? If you are using the 2 RE node system, there must either be some individuals comprising >2 recombination nodes, or there must be some recombination nodes associated with >1 recombination?

The *better* way to think about it, if you don't want to call a recombination node a gamete, is to call the *child* node the recombination node, and store the breakpoint there (if, indeed, you want to store the breakpoint at all).

I was going to put together a little schematic to try to explain all this. I *strongly* think that we should use a single node for the recombination node, for this reason. But again, maybe easier to explain in person.

jeromekelleher:
I've given up on trying to make the non-annotated edge graph work @hyanwong - it basically doesn't. It only happens to work in the limited case where the coalescent approximations hold, and trying to apply it elsewhere is a waste of time, IMO. All we're doing in coming up with these elaborate workarounds is making things more confusing.

hyanwong:
> I've given up on trying to make the non-edge-annotated graph work @hyanwong - it basically doesn't. It only happens to work in the limited case where the coalescent approximations hold, and trying to apply it elsewhere is a waste of time, IMO. All we're doing in coming up with these elaborate workarounds is making things more confusing.

I partially agree with this. I can see a (small) benefit to encoding the breakpoint in the metadata for general reference, but it shouldn't be seen as any sort of alternative to storing the data on edges. But I'm not terribly bothered about it either way: it's a nicety, that's all.

I still think that the 2-RE-node scheme is wrong, though, for a more fundamental biological reason: I think each RE node (or pair of Re nodes) should describe a *single* meiosis (i.e. a single round of recombination). If you have 2 nodes (and imagine that those 2 nodes represent the genomes in the diploid parent), then in a WF model, because a parent can produce multiple recombinations, one in each gamete, the 2 nodes stand in for multiple independent recombinations in different cell lineages. That's the picture I wanted to produce to illustrate. I'll see if I can throw something together tomorrow, but perhaps you get the point anyway?

Alternatively, you have >2 RE nodes per individual, but that breaks the tskit model.

hyanwong:
I had a look through the code, but I think it would be easier to me to grasp if you explained it. Perhaps at 11 today?

By the way, another way to think about the RE node thing is simply labelling. Labelling the 2 nodes that come together (forwards in time) to create a gamete as the "recombination nodes" means that a RE label means "a genome that recombines, *one or more times*, with the other genome in this pair, to create a recombined gamete in the next generation". Labelling the gamete would mean that a RE node is "a genome in a gamete that is the result of a single round of recombination", and labelling the *child* node as the RE node would mean "a genome in an individual that was created by a single round of recombination from the 2 parents in the previous generation". As I have been arguing, I prefer the last 2 meanings of the label to the first one, but all are accurate labels.

I'm not quite sure, if we don't care about breakpoints in non-ancestral material, why we think that the 2 node version is generally preferable? Contrary to the title of https://github.com/tskit-dev/what-is-an-arg-paper/discussions/41, I don't think that the problem with the Griffiths graph is with the 1-node approach per se: it's with the node annotation. We shouldn't use a 2 node model by default just because it rescues the node-encoded version of a Griffiths graph. In fact, we shouldn't use requirements of the CwR to decide the general encoding for an ARG structure.

jeromekelleher:
Thanks, this is helpful. One thing I'm confused about though: why don't you want one node to correspond to an individual's genome? As in, each individual consists of two genomes/nodes. We have (annotated) edges between those genomes if there is genetic inheritance. That's the end of it then, surely: there's no need to invoke gametes or partial generations or any of the rest of it?

jeromekelleher:
After our discussion earlier @hyanwong I've updated the code to move the RE flag to the child node. This actually simplifies things quite a bit in terms of reasoning about flags.

I also added some metadata to record the CA status of the nodes, as they are being created. Here's the output of the current script:

```

20.00┊                  29:1/1         ┊                                 ┊                                 ┊                                 ┊
     ┊              ┏━━━━━━┻━━━━━━┓    ┊                                 ┊                                 ┊                                 ┊
11.00┊              ┃             ┃    ┊                                 ┊                 28:1/1          ┊                                 ┊
     ┊              ┃             ┃    ┊                                 ┊            ┏━━━━━━━┻━━━━━━━┓    ┊                                 ┊
10.00┊           25:0/1        27:0/1  ┊                                 ┊         24:0/1          26:0/1  ┊                                 ┊
     ┊              ┃             ┃    ┊                                 ┊            ┃               ┃    ┊                                 ┊
9.00 ┊          R22:0/2       R23:0/2  ┊                                 ┊        R22:0/2         R23:0/2  ┊                                 ┊
     ┊              ┃             ┃    ┊                                 ┊            ┃               ┃    ┊                                 ┊
8.00 ┊          R20:0/2       R21:0/2  ┊                                 ┊        R21:0/2         R20:0/2  ┊                                 ┊
     ┊              ┃             ┃    ┊                                 ┊            ┃               ┃    ┊                                 ┊
6.00 ┊           18:0/1        17:0/1  ┊                                 ┊         16:0/1          19:0/1  ┊                                 ┊
     ┊              ┃             ┃    ┊                                 ┊            ┃               ┃    ┊                                 ┊
5.00 ┊              ┃         R15:0/2  ┊                                 ┊        R15:0/2             ┃    ┊                                 ┊
     ┊              ┃             ┃    ┊                                 ┊            ┃               ┃    ┊                                 ┊
4.00 ┊          R13:0/2        14:0/1  ┊                 12:1/2          ┊         12:1/2         R13:0/2  ┊                                 ┊
     ┊              ┃             ┃    ┊            ┏━━━━━━━┻━━━━━━━┓    ┊            ┃               ┃    ┊                                 ┊
3.00 ┊          R11:1/3        10:0/1  ┊        R11:1/3          R9:1/3  ┊        R11:1/3          R9:1/3  ┊             R9:1/3              ┊
     ┊        ┏━━━━━┻━━━━━┓       ┃    ┊            ┃               ┃    ┊            ┃               ┃    ┊        ┏━━━━━━━┻━━━━━━━┓        ┊
2.00 ┊        ┃           ┃    R8:1/4  ┊            ┃            R8:1/4  ┊            ┃            R8:1/4  ┊        ┃            R8:1/4      ┊
     ┊        ┃           ┃       ┃    ┊            ┃               ┃    ┊            ┃               ┃    ┊        ┃           ┏━━━┻━━━┓    ┊
1.00 ┊     R5:1/2      R4:3/4  7:0/1   ┊         R4:3/4          6:0/3   ┊         R4:3/4          6:0/3   ┊     R4:3/4      R5:1/2  6:0/3   ┊
     ┊    ┏━━━┻━━━┓       ┃       ┃    ┊    ┏━━━━━━━╋━━━━━━━┓       ┃    ┊    ┏━━━━━━━╋━━━━━━━┓       ┃    ┊    ┏━━━┻━━━┓       ┃       ┃    ┊
0.00 ┊ R0:4/4  R2:4/4  R1:4/4  R3:4/4  ┊ R0:4/4  R1:4/4  R2:4/4  R3:4/4  ┊ R0:4/4  R1:4/4  R2:4/4  R3:4/4  ┊ R0:4/4  R2:4/4  R1:4/4  R3:4/4  ┊
     0                                 1                                 2                                 3                                 4
```

The annotations on the node labels are an initial "R" if it is a recombination node. The bit after the ":" is then the ``coal_span`` and  the ``total_span``.

The ``total_span`` is the the sum of the spans of the intervals over which the corresponding ancestral genome carries ancestral material (note this is *not* the ``right - left`` of the overall span, as this would include non ancestral trapped material). The ``coal_span`` is then the sum of the intervals over which some coalescence occured.

So, node 5 is a recombinanant, and it has an ancestral span of 2 bases, of which it is a coalescent over 1 base. Note 18 (say) is a boring "nothing to see here" node that it is very difficult to imagine ever being detectable.

Is this useful to do you think?

jeromekelleher:
(Note I haven't updated the standard CwR simulations to move the RE flags to the child; I'll see how this looks later)

jeromekelleher:
I've added the RE flag to the child node in the CwR simulator @hyanwong, and some interesting questions pop up. Consider these ARGs:
```
2.66┊       13     ┊     13      ┊             ┊
    ┊     ┏━━┻━━┓  ┊   ┏━━┻━━┓   ┊             ┊
2.14┊    12     ┃  ┊   ┃    12   ┊     12      ┊
    ┊     ┃     ┃  ┊   ┃     ┃   ┊   ┏━━┻━━┓   ┊
1.48┊     ┃    R11 ┊  R11    ┃   ┊  R11    ┃   ┊
    ┊     ┃     ┃  ┊  ┏┻━┓   ┃   ┊  ┏┻━┓   ┃   ┊
1.16┊    10     ┃  ┊  ┃  ┃  10   ┊  ┃  ┃  10   ┊
    ┊     ┃     ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
0.86┊     9     ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
    ┊  ┏━━┻━┓   ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
0.55┊ R8    ┃   ┃  ┊ R8  ┃   ┃   ┊ R8  ┃   ┃   ┊
    ┊ ┏┻┓   ┃   ┃  ┊ ┏┻┓ ┃   ┃   ┊ ┏┻┓ ┃   ┃   ┊
0.06┊ ┃ ┃  R7   ┃  ┊ ┃ ┃ ┃  R7   ┊ ┃ ┃ ┃  R7   ┊
    ┊ ┃ ┃ ┏━┻┓  ┃  ┊ ┃ ┃ ┃ ┏━┻┓  ┊ ┃ ┃ ┃ ┏━┻┓  ┊
0.03┊ ┃ ┃ ┃  6  ┃  ┊ ┃ ┃ ┃ ┃  6  ┊ ┃ ┃ ┃ ┃  6  ┊
    ┊ ┃ ┃ ┃ ┏┻┓ ┃  ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊
0.00┊ 0 2 1 4 5 3  ┊ 0 2 3 1 4 5 ┊ 0 2 3 1 4 5 ┊
    0              1             2             5

2.66┊       19     ┊     19      ┊             ┊
    ┊     ┏━━┻━━┓  ┊   ┏━━┻━━┓   ┊             ┊
2.14┊    18     ┃  ┊   ┃    18   ┊     18      ┊
    ┊     ┃     ┃  ┊   ┃     ┃   ┊   ┏━━┻━━┓   ┊
2.00┊     ┃    16  ┊  16     ┃   ┊  17     ┃   ┊
    ┊     ┃     ┃  ┊   ┃     ┃   ┊   ┃     ┃   ┊
1.48┊     ┃    R15 ┊  R15    ┃   ┊  R15    ┃   ┊
    ┊     ┃     ┃  ┊  ┏┻━┓   ┃   ┊  ┏┻━┓   ┃   ┊
1.16┊    14     ┃  ┊  ┃  ┃  14   ┊  ┃  ┃  14   ┊
    ┊     ┃     ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
0.86┊    13     ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
    ┊  ┏━━┻━┓   ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
0.74┊  ┃   11   ┃  ┊  ┃  ┃  12   ┊  ┃  ┃  12   ┊
    ┊  ┃    ┃   ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
0.61┊  9    ┃   ┃  ┊ 10  ┃   ┃   ┊ 10  ┃   ┃   ┊
    ┊  ┃    ┃   ┃  ┊  ┃  ┃   ┃   ┊  ┃  ┃   ┃   ┊
0.55┊ R8    ┃   ┃  ┊ R8  ┃   ┃   ┊ R8  ┃   ┃   ┊
    ┊ ┏┻┓   ┃   ┃  ┊ ┏┻┓ ┃   ┃   ┊ ┏┻┓ ┃   ┃   ┊
0.06┊ ┃ ┃  R7   ┃  ┊ ┃ ┃ ┃  R7   ┊ ┃ ┃ ┃  R7   ┊
    ┊ ┃ ┃ ┏━┻┓  ┃  ┊ ┃ ┃ ┃ ┏━┻┓  ┊ ┃ ┃ ┃ ┏━┻┓  ┊
0.03┊ ┃ ┃ ┃  6  ┃  ┊ ┃ ┃ ┃ ┃  6  ┊ ┃ ┃ ┃ ┃  6  ┊
    ┊ ┃ ┃ ┃ ┏┻┓ ┃  ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊ ┃ ┃ ┃ ┃ ┏┻┓ ┊
0.00┊ 0 2 1 4 5 3  ┊ 0 2 3 1 4 5 ┊ 0 2 3 1 4 5 ┊
    0              1             2             5
```

They are identical realisations of the coalescent, but one is storing the parent nodes for recombination events and the other isn't. Given that we're now labelling the child as the recombinant and associating the breakpoint with it, we don't actually need to store the parent nodes to know exactly where and on what lineages the recombination happened. We do lose the time of the event though.

It's probably pretty philosophical whether we regard these parent nodes as part of the ARG or not

jeromekelleher:
> Great, thanks @jeromekelleher. Do you want to merge this before I try the viz stuff?

OK, merging now.

The ``resolved_wf_arg_sim`` and ``arg_sim`` functions are up to date - other's may not work as expected.

hyanwong:
> Given that we're now labelling the child as the recombinant and associating the breakpoint with it, we don't actually need to store the parent nodes to know exactly where and on what lineages the recombination happened. 

Yes, this is the nice thing. Those nodes turn out to be pass-though, and so can often be discarded.

The standard "time" of the recombination must be 1 generation before the RE node, by definition, I think. But in fact, the date of "birth" (or conception) of the child is, for sperm, much closer to the time of recombination than the time of birth of the parent node. For eggs in mammals, the time is probably closer to the birth time of the parent through. It's all an approximation anyway.

jeromekelleher:
> It's probably pretty philosophical whether we regard these parent nodes as part of the ARG or not

Aha, no, I realised I'm wrong here. We **must** stored the parent nodes or we can have multiple recombinations stacking up on a node and we lose information about the process. I'll push an update PR

hyanwong:
> > It's probably pretty philosophical whether we regard these parent nodes as part of the ARG or not
> 
> Aha, no, I realised I'm wrong here. We **must** stored the parent nodes or we can have multiple recombinations stacking up on a node and we lose information about the process. I'll push an update PR

I don't think we can have multiple recombinations stacking up on a recombination node, can we?

jeromekelleher:
> The standard "time" of the recombination must be 1 generation before the RE node, by definition, I think. But in fact, the date of "birth" (or conception) of the child is, for sperm, much closer to the time of recombination than the time of birth of the parent node. For eggs in mammals, the time is probably closer to the birth time of the parent through. It's all an approximation anyway.

Not in this model, no. Look at the distance between R7 and its parents 11 and 12.

jeromekelleher:
> I don't think we can have multiple recombinations stacking up on a (child) recombination node, can we?

We can if we don't change the node associated with the lineages (by moving to the parents). A quick assert in the code shows it happens pretty quickly that you try to add a breakpoint to node metadata that already has a breakpoint.

hyanwong:
> Not in this model, no. Look at the distance between R7 and its parents 11 and 12.

Oh, this isn't the WF model, but the CwR (basically Moran) model? Well, the same thing applies though. The date of birth of the RE node is usually close in time to the production of the gametes that made it, which is the "true" time of the recombination, right?

PR #46, opened 2022-01-30
New abstract + introduction, start of ARG explanation section

jeromekelleher:
This is a WIP, but might be worth merging if others want to work on the text.

Intro and abstract text are in a "getting there" state I think. Clumsy phrasing etc, but I think the narrative is starting to come together.

Any thoughts @JereKoskela @hyanwong?

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -66,7 +66,7 @@ and discuss the properties of some recently developed inference methods.
 \section*{Introduction}
 Inferring the genetic relationships between sampled genomes in the form of an
 evolutionary tree is a necessary prerequisite for many analyses of non (or
-rarely) recombining organisms [refs]. The trees capture what is known and
+rarely) recombining organisms [refs]. The tree captures what is known and
 knowable about the genetic ancestry of the sample in an elegant and concise
 way, by postulating a set of nodes that are ancestors common to the samples,
 and the ancestral relationships between those nodes. A rich literature exists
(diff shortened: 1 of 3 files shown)


paper.tex
```diff
@@ -22,46 +26,256 @@
...
+% how the ARG relates the pedigree, etc, etc. Let's go through again when
+% we have a reasonable draft.
 \end{abstract}
 
 \textbf{Keywords:} Ancestral Recombination Graphs
 
 \section*{Introduction}
+Inferring the genetic relationships between sampled genomes in the form of an
+evolutionary tree is a necessary prerequisite for many analyses of non (or
+rarely) recombining organisms [refs]. The trees capture what is known and
```
JereKoskela, line 69: Minor nit-pick, but should "The trees capture" be the singular "The tree captures" here, to avoid any insinuation of multiple trees for the same non-recombining sample?

JereKoskela:
This looks great @jeromekelleher, thanks! I think having these early sections pinned down in such a clear shape will make it much easier to keep later drafts terminologically and conceptually consistent.

paper.tex
jeromekelleher: Sure, lets go with that

jeromekelleher:
I'm going to merge this much as the early sections are fairly OK, and I don't want my version to diverge too much from the repo.

PR #50, opened 2022-02-01
A few minor language tweaks

hyanwong:


paper.tex
```diff
@@ -41,15 +41,15 @@
...
-stochastic process as a graph, an ARG is now understood to refer to any
+a stochastic process (the coalescent with recombination) together with its 
+representation, an ARG is now understood to refer to any
 concrete realisation of a recombinant genetic ancestry.
 We show that the standard description of an ARG
 in terms of common ancestor and recombination events
 (inherited from the original stochastic process definition) is
 ambiguous, fundamentally limited and an unsuitable basis for data interchange.
 We provide a simple alternative definition of an ARG in which a node
-corresponds to one of an individuals monoploid genomes, an edge
```
jeromekelleher, line 52: I think we need to be ultra clear and specific about this though, even if it is technically redundant. People will skip over it without thinking it through properly if we just say "a genome"

paper.tex
```diff
@@ -73,9 +73,9 @@ \section*{Introduction}
 on analysing the mathematical properties of these trees [cite?], and numerous
 methods exist to infer them~\citep{felsenstein2004inferring}. The situation is
 quite different in recombining organisms, where the analysis of genetic data
-depends on summaries ultimately derived from the trees such as allele
+depends on summaries ultimately derived from the trees~\citep{ralph2020efficiently}
 % TODO cite the other papers that preceded ralph2020efficiently here.
-frequencies~\citep{ralph2020efficiently}, site frequency spectra, patterns of linkage
```
jeromekelleher, line 78: This positioning of the citation was intentional: all of these things are references to papers that go through the equivalance of that particular thing with trees. Putting this first makes it look like ralph2020efficiently explains things like PCA etc, which it doesn't.

paper.tex
```diff
@@ -106,8 +106,9 @@ \section*{Introduction}
 and show how it derives from (and is limited by) the coalescent with
 recombination.
 \item Suggest a definition of ARGs that is free of these limitations,
-where the formulation is derived from a diploid pedigree rather than
-the coalescent approximations and spatial annotations are attached to
+where the formulation is derived from a pedigree rather than from
```
jeromekelleher, line 109: Again, I think we have to be specific here or people won't think it through properly. Putting "diploid" in there makes people think about the multiple genomes explicitly. I know there's no particular reliance of a given ploidy level, but that's something we can clarify later.

Genetic material is too vague as well, we need something that gets across the idea that we're talking about the actual genome coordinates.

paper.tex
```diff
@@ -160,14 +161,14 @@ \section*{Ancestral graphs and stochastic processes}
...
-tracking the number of samples the lineage is ancestral to over the interval.
+tracking the number of samples to which the lineage is ancestral over that interval.
 (We also usually track the tree node associated with each segment, but
 that is not important for our purposes here so we omit it.)
 If we have $n$ samples and genome of length $m$, the process begins with $n$ lineages
 of the form $[(0, m, 1)]$. The process then works backwards in time from
 the present day as a series of random common ancestor or recombination events.
-Recombination events occur at a rate determined by the amount and spatial
-distribution of ancestral material among lineages.
+Recombination events occur at a rate determined by the genetic location and amount
```
jeromekelleher, line 170: As Jere suggests, "at a rate determined by the amount of ancestral material and the way in which it is distributed along a chromosome"

"Genetic location" suggests we're talking about genetic maps or something, which is just an extra layer of confusion to avoid. (I mean, we are, but it's not important)

paper.tex
```diff
@@ -344,8 +345,8 @@ \section*{The ARG as a data structure}
 the data structure interpretation.
 
 The Griffiths ARG encoding is an elegant and mathematically
-economical description of the coalescent with recombination, but
-it has significant drawbacks as the basis for a general purpose
+economical description of the coalescent with recombination on a single
```
jeromekelleher, line 348: This implies we're talking about a sample of size 1 to me - what are you trying to clarify here?

jeromekelleher (commented):
Thanks @hyanwong - some useful updates here. Also some regressions though I think - I'd rather have some redundancy in the language here if it gives any increase at all in the likelihood people will understand us. There's huge room for confusion here, particularly among those who are likely to be skim reading this.

paper.tex
hyanwong: That's a fair point. I wanted to make sure that it wasn't just diploids that this applied to, but your previous phrasing didn't state that, so I guess that's OK. It just needs the apostrophe in there then.

paper.tex
hyanwong: Ah, I see. I moved it because I feel like we need a citation that to back up the "summaries ultimately derived from the trees" quote, and the "duality" paper applies to the SFS too. But I'll put it back and we can discuss. Is there a paper out there that makes the point about the summaries of whatever sort being derived from trees?

paper.tex
hyanwong: The formulation need not be derived from a standard diploid pedigree, though? It would be just as reasonable to derived it from e.g. a gametophytic system, I think. So we just want to be careful here not to tie it too strongly to the human-style pedigree, I think?

paper.tex
hyanwong: Oh, I see. I mean that the Griffiths model only applies to an organism with a single chromosome. That's actually a much greater restriction than the one about multiple breakpoints, because most (all?) organisms have multiple chromosomes, so the inability to represent these is a significant deviation from the ability to describe the real world.

hyanwong:
> Thanks @hyanwong - some useful updates here. Also some regressions though I think -

Yep, agree, this is a first pass, for comments, so thanks for those.

paper.tex
jeromekelleher: > Is there a paper out there that makes the point about the summaries of whatever sort being derived from trees?

Not that I know of off the top of my head, but it'll be referred to in the stats paper if it does. If you're looking, it would be helpful to get the other papers talking about allele frequencies and trees from the references, like the comment mentions.

paper.tex
jeromekelleher: As I say, I know there's no strong dependence on particular ploidy levels here. This isn't a mature draft and I never even remotely thought of this text as being final.

jeromekelleher:
OK, thanks. Can you fix the regressions so please and we can merge?

paper.tex
hyanwong: Perhaps we just merge the freq + SFS items:

```
where the analysis of genetic data
depends on summaries ultimately derived from the trees
such as allele frequencies and site frequency spectra ~\citep{ralph2020efficiently}, patterns of linkage
disequilibrium~\citep{mcvean2002genealogical} and principal
components~\citep{mcvean2009genealogical}
```

paper.tex
hyanwong: Yep, absolutely. I know this is a draft section, I just thought it worth changing the language so that it didn't end up going into the final text by default/accident. But happy to revery and reconsider when the final text is done.

paper.tex
hyanwong: > Not that I know of off the top of my head, but it'll be referred to in the stats paper if it does. If you're looking, it would be helpful to get the other papers talking about allele frequencies and trees from the references, like the comment mentions.

The relevant section of the stats paper says this:
> This duality (we will use the term more precisely later) has been used extensively in deriving properties of statistics in models of random mating (e.g., Tajima 1983; Tavaré 1984; Fu 1995), as has the fact that it applies regardless of the underlying demographic model (e.g., Gillespie and Langley 1979; Hudson 1983; Slatkin 1991; McVean 2002; Lohse et al. 2016; Ralph 2019)

jeromekelleher: These are the ones I was thinking of:

> Suppose we want to calculate one of the linear functions of the frequency spectrum discussed by Fu (1995) and Achaz (2009),

hyanwong: There's also [Felsenstein 1992](https://www.cambridge.org/core/journals/genetics-research/article/estimating-effective-population-size-from-samples-of-sequences-inefficiency-of-pairwise-and-segregating-sites-as-compared-to-phylogenetic-estimates/C199D5519B61D02C7052EAB262FA7449)

jeromekelleher:
Can you fix these regressions please @hyanwong? I wanted to do some writing here, but with changes scattered all over the document there's going to be lots of conflicts.

hyanwong:
> Can you fix these regressions please @hyanwong? I wanted to do some writing here, but with changes scattered all over the document there's going to be lots of conflicts.

Just looking up the refs. Will fix now and go back to it

hyanwong:
> Can you fix these regressions please @hyanwong? I wanted to do some writing here, but with changes scattered all over the document there's going to be lots of conflicts.

Done and merged

paper.tex
hyanwong: The (Tajima 1983; Tavaré 1984; Fu 1995) cites are the right ones here. E.g. from Tavaré:

> The purpose of this paper is to review some of the pertinent mathematical details about genealogical processes, lines of descent, and their structure and demonstrate how such ideas can be applied to produce a wide variety of classical results in the mathematical theory of population genetics.

PR #57, opened 2022-02-02
Modified first sentence, added references

awohns:
I think the first sentence can be modified to better describe how evolutionary trees are practically used. I see the main categories of tree inference as: (1) species trees, (2) trees of non-recombining sections of genome, and (3) viral phylogenies. Do others like this classification?
Another reason to rephrase: we can't really cite viral phylogenetics papers with the current wording since viruses aren't usually referred to as "organisms".

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -998,3 +998,45 @@
   year={2016},
   publisher={SIAM}
 }
+
+
+@article{rannala2003genetics,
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Looks great, thanks. Can you fix the conflicts please?

PR #59, opened 2022-02-04
Consolidate code into package

jeromekelleher:
This is a WIP so letting you know it's happening @hyanwong @awohns 

The idea is to consolidate the various bits of scripts into a package (argutils) so that we can import it in the normal way rather than faddling around with the Python path. I've also put in a stub for a CLI, which we can hopefully use to do some of the things we need rather than distributing code around in various scripts.

You'd use it like ``python -m argutils <command> <args>``

I'm putting in some basic tests for the fundamental stuff, but there'll be no expectation for tests on things like visualisations.


hyanwong:
Good call. Thanks @jeromekelleher

jeromekelleher:
I'm going to merge this as some of the code is in pretty good shape. There's still a bunch of stuff quite raw, but it's a good starting point (and there's no point in working with existing code).

PR #63, opened 2022-02-06
Refine argutils

jeromekelleher:
Two new things here:

1. It turns out that the resolve algorithm *is* actually simplify(keep_unary=True). I had been confused about what happens over the roots, but #61 has shown that simplify is actually doing the right thing. I haven't been able to find a counter example where it doesn't, but it must be right.
2. Put some simple infrastructure in for playing with visualisation by porting some of Yan's code. Now, running 

```
python3 -m argutils draw tmp.trees tmp.png
```
will draw a simple plot to the file. Currently it looks like this:

tmp

The layout looks good (although, not clear why node 4 is at the same y value as 7 and 8 - the time is different), but there's clearly plenty that could be done. The edge information is stored in the nx Graph object, so that should be sufficient to start working out what we want to display.

Look OK as a starting point @hyanwong, @awohns?

### DIFF
--- a/argutils/__init__.py
+++ b/argutils/__init__.py
@@ -2,4 +2,4 @@
 Various utilities for working with args.
 """
 from . ancestry import *
-from . viz import *
+from . viz import draw
--- a/argutils/__main__.py
+++ b/argutils/__main__.py
@@ -1,5 +1,7 @@
 import click
 import argutils
+import tskit
+import matplotlib.pyplot as plt
 
 
 @click.group()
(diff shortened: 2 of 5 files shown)


hyanwong:
> It turns out that the resolve algorithm is actually simplify(keep_unary=True)

By the "resolve" algorithm you mean the conversion of the "local" edge annotated ARG (AKA Griffiths edge-annotated ARG) to the tree sequence?

What happens if we have a recombinant node that is not a coalescent node anywhere? I guess we keep these nodes so that we can change it back into the "local" version?

jeromekelleher:
> By the "resolve" algorithm you mean the conversion of the "local" edge annotated ARG (AKA Griffiths implicit ARG) to the explicit tree sequence?

Yes - sorry, the terminology keeps changing here. It's the process of turning an ARG with local inheritance information only into one that's fully resolved WRT to the ancestral material of the samples.

> What happens if we have a recombinant node that is not a coalescent node anywhere? I guess we keep these nodes so that we can change it back into the "local" version?

I think they should be kept - we don't want to throw away ARG nodes in this process, we just want to be explicit about how much (and where) of the genome that's being transmitted along the edge is ancestral to the samples.

jeromekelleher:
Going to merge this as a good start to work on.

PR #67, opened 2022-02-07
Added "back in time" and "along the genome" etc descriptions of parsimony/heuristic methods

a-ignatieva:
Draft wording, as discussed in #14


hyanwong:
This reads really well @a-ignatieva. I think we should just merge it an iterate if necessary.

jeromekelleher:
I agree, merging! Thanks @a-ignatieva

Issue #65, opened 2022-02-07
Layout nodes by time in viz


jeromekelleher:
The current ``argutils.draw()`` method produces slightly odd layouts in the y direction. See here for example:

![wh99](https://user-images.githubusercontent.com/2664569/152805107-bed4d497-57cf-474d-a295-913dc3ffd4b3.png)

It would be nice to be able to layout the nodes in time-rank order so that here nodes 3 and 4 (which have the same time) are at the same level, but different to all the others.

@hyanwong - is this possible with the dot layout used for the x layout (which is great)?

hyanwong:
There is supposedly a way to do it in dot, but I've never got it to work satisfactorily. There's a comment where we say "# First cluster all nodes at the same times (probably mostly samples)". OTOH we do change the Y coords from the dot layout after the fact. Let me see what I can do.

jeromekelleher:
The ``pos = nx.multipartite_layout(G, subset_key="time", align="horizontal") `` does a good job of laying out by time for pedigrees.

hyanwong:
The nx layouts don't try to minimise line crossing at all, which is why I had to use dot. We could then use the Y positions from the multipartite layout, I suppose.

hyanwong:
Just had a look. That's what the "add invisibles" thing does. Shall I expose it in the `draw` function?

It doesn't always work as expected, which is why I didn't turn it on be default. The resulting plot does weird stuff in the X direction now:

![Screenshot 2022-02-08 at 09 41 17](https://user-images.githubusercontent.com/4699014/152960141-3c29c4d1-e29c-4342-b36e-f19634c3d07c.png)

jeromekelleher:
That does look worse layout wise now. I think it'd probably be simpler to do the y layout ourselves - what we want is straightforward isn't it?

hyanwong:
Yep, I think we just impose the ranks onto the Y axis ourselves.

hyanwong:
![Screenshot 2022-02-08 at 10 12 10](https://user-images.githubusercontent.com/4699014/152966179-6943e649-7d93-4e13-85eb-c817ba73bf72.png)

Using

```
def draw(ts, ax, use_ranked_times=None):
    """
    If use_ranked times is True, the y axis uses the time ranks, with the
    same times sharing a rank. If False, it uses the true (tree sequence)
    times. If None, times from the tree sequence are not used and the
    standard dot layout is used
    """
    G = convert_nx(ts)
[...]

PR #69, opened 2022-02-08
When drawing ARGs, allow y times to be used and hand tweak X pos

hyanwong:
A bit hacky, but I suspect some hand-tweaking ability would be generally useful. For instance:

```
import argutils

tables = argutils.wh99_example()
tables.sort()
ts = tables.tree_sequence()

fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10,10))
ax1.set_title("dot layout without X tweaks")
argutils.viz.draw(ts, ax1)
ax2.set_title("true times, tweaking X for nodes 0 and 2")
argutils.viz.draw(ts, ax2, use_ranked_times=False, tweak_x={0: -10, 2: 10})
```
Screenshot 2022-02-08 at 10 57 11

### DIFF
--- a/argutils/viz.py
+++ b/argutils/viz.py
@@ -8,10 +8,34 @@ import networkx as nx
 import numpy as np
 
 
-def draw(ts, ax):
+def draw(ts, ax, use_ranked_times=None, tweak_x=None):
+    """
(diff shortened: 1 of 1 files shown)


hyanwong:
Closes #65

jeromekelleher:
LGTM

PR #71, opened 2022-02-08
Make the returned object a directed graph

hyanwong:
But don't draw arrows by default. Drawing edges separately will also make it easier to loop over multiple edges in a simplified TS, if we end up needing that.


jeromekelleher:
LGTM

PR #72, opened 2022-02-08
Add ability to label nodes

hyanwong:
Use metadata labels to make nodes labels comparable even after simplification. As per other figures, we default to ascii uppercase letters. Use like this:

```
import argutils

ts = argutils.viz.label_nodes(argutils.sim_coalescent(4, 0.001, 1000, seed=123))
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(15,10))
argutils.viz.draw(ts, ax1, use_ranked_times=None)
argutils.viz.draw(ts.simplify(), ax2, use_ranked_times=None)
```
image

### DIFF
--- a/argutils/viz.py
+++ b/argutils/viz.py
@@ -6,7 +6,7 @@ import collections
 import tskit
 import networkx as nx
 import numpy as np
-
+import string
 
 def draw(ts, ax, use_ranked_times=None, tweak_x=None, arrows=False):
     """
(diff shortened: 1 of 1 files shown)


argutils/viz.py
```diff
@@ -58,6 +68,33 @@ def draw(ts, ax, use_ranked_times=None, tweak_x=None, arrows=False):
...
+    simply have their metadata value set to their node id.
+    
+    Note that this means that if no labels are given, nodes 26 onwards
+    will be labelled with numbers rather than ascii uppercase letters.
+    """
+    if labels is None:
+        labels = {i: lab for i, lab in enumerate(string.ascii_uppercase)}
+    tables = ts.dump_tables()
+    try:
+        if tables.nodes.metadata_schema.asdict()["codec"] != "json":
```
jeromekelleher, line 86: Probably not worth bothering with this - it'll probably end up being as fragile as just going ahead and assuming the codec is JSON.

argutils/viz.py
```diff
@@ -58,6 +68,33 @@ def draw(ts, ax, use_ranked_times=None, tweak_x=None, arrows=False):
     )
 
 
+def label_nodes(ts, labels=None, metadata_name="name"):
```
jeromekelleher, line 71: I can't see any reason why we'd change the name of the attr - why add all these extra parameters for a simple utility that we're only using in this paper? Let's keep it simple...

jeromekelleher (approved):
LGTM - but let's try and keep the number of parameters going into these functions under control, and only add knobs to tweak things we actively need to tweak.

jeromekelleher:
Great idea btw - really makes it easy to keep track of the nodes.

argutils/viz.py
hyanwong: Fair point. Are we happy with sticking with "name" then?

argutils/viz.py
hyanwong: Should I wrap the metadata setting code in a try: except block. I was worried that it would be a confusing error if run on a ts without json metadata.

hyanwong:
OK, fixed. Should be OK to merge.

hyanwong:
Merging, because I have some other changes to layer on top of this, and comments have been addressed

jeromekelleher:
Yep, looks great

PR #73, opened 2022-02-08
Remove resolve() method and return tree sequence

jeromekelleher:
Separate resolve() method is superfluous, and it's more convenient to return a tree sequence when ``resolve=False``.


hyanwong:
LGTM. Great that `simplify` does it all for us! I was going to ask about returning a ts rather than the tables too. Glad you beat me to it. Just merge this, I think.

PR #74, opened 2022-02-08
Added Hudson vs big ARG figure

### LINKED ISSUE #55: Figure 1: Hudson's alg vs Big ARG
We'd like a figure showing the differences in the state space between Hudson's algorithm and the Big ARG. Hudson's algorithm should show the full state, include the (left, right, ancestral_to) tuples that are associated with each lineage.

The ``arg_sim`` function might be helpful for inspiration [here](https://github.com/tskit-dev/what-is-an-arg-paper/blob/7fb73720d63e930c25c29068367c428db069adf1/utils/args.py#L203)

@JereKoskela - are you happy picking this one up?

JereKoskela:
Closes #55. We can always reopen a more targeted issue if we want to discuss polishing this.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -7,6 +7,7 @@
 \usepackage{authblk}
 \usepackage{nameref}
 \usepackage[right]{lineno}
+\usepackage{tikz}
 
 \usepackage{subcaption}
 \usepackage{tikz}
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Awesome, merging!

Issue #75, opened 2022-02-08
Graph viz layout seemingly non-deterministic


hyanwong:
Weirdly, when I run the following code after restarting the notebook kernel, I get slightly different X layouts. This is *really* annoying. It seems like the dot layout engine should be deterministic, so I don't know what's going on here.

```
ts = argutils.sim_coalescent(4, 0.001, 1000, seed=123)
print(f"{ts.num_samples} samples, {ts.num_trees} trees")
fig, (ax1) = plt.subplots(1, 1, figsize=(10,10))
argutils.viz.draw(ts, ax1, use_ranked_times=True)
```

1st attempt
![image](https://user-images.githubusercontent.com/4699014/153031415-39a23221-d559-42a4-b646-54077701630d.png)

repeat attempt
![image](https://user-images.githubusercontent.com/4699014/153031459-b7687c10-0be9-4cc3-8c6f-00256dae4c4a.png)

jeromekelleher:
That is very irritating. Do we know what the underlying calls to graphviz are? We could do a bit of digging then.

hyanwong:
Looks like the returned values from `A.get_node(n).attr["pos"]` are indeed different:

```
1st iteration      2nd iteration
node 0: 301,738    node 0: 193,738 
node 1: 85,738     node 1: 121,738 
node 2: 229,738    node 2: 337,738 
node 3: 157,738    node 3: 265,738 
node 4: 130,522    node 4: 115,666 
node 5: 58,522     node 5: 43,666  
node 6: 197,666    node 6: 337,666 
node 7: 91,450     node 7: 79,594  
node 8: 352,666    node 8: 265,666 
node 9: 280,666    node 9: 193,666 
node 10: 331,59    node 10: 195,5944
node 11: 259,59    node 11: 267,5944
node 12: 187,59    node 12: 339,5944
node 13: 259,52    node 13: 267,5222
node 14: 211,45    node 14: 272,4500
node 15: 43,378    node 15: 53,378 
node 16: 115,37    node 16: 125,3788
node 17: 265,37    node 17: 275,3788
node 18: 193,37    node 18: 203,3788
node 19: 187,30    node 19: 197,3066
node 20: 187,23    node 20: 197,2344
node 21: 187,16    node 21: 197,1622
node 22: 115,16    node 22: 125,1622
node 23: 115,90    node 23: 125,90 
node 24: 142,18    node 24: 152,18 
```

[...]

PR #78, opened 2022-02-08
Return positions from draw, and allow positions to be specified

hyanwong:
This allows us to place nodes in the same position in adjacent plots. It could also act as a workaround for #75 as we could output the positions and then hard code them. Here's how we could do the former:

```
ts = argutils.sim_coalescent(4, 0.001, 1000, seed=123)
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(20,10))
pos = argutils.viz.draw(ts, ax1, use_ranked_times=True, tweak_x={24: 15, 23: 5, 22: 10, 21: 10})
new_ts, node_map = ts.simplify(map_nodes=True)
_ = argutils.viz.draw(new_ts, ax2, pos={node_map[i]:p  for i, p in pos.items()})
```
image

### DIFF
--- a/argutils/viz.py
+++ b/argutils/viz.py
@@ -8,8 +8,12 @@ import networkx as nx
 import numpy as np
 import string
 
-def draw(ts, ax, use_ranked_times=None, tweak_x=None, arrows=False):
+def draw(ts, ax, use_ranked_times=None, tweak_x=None, arrows=False, pos=None):
     """
+    Draw a graphical representation of a tree sequence. If a metadata key
+    called "name" exists for the node, it is taken as
+    a node label, otherwise the node ID will be used as a label instead.
+
     If use_ranked times is True, the y axis uses the time ranks, with the
     same times sharing a rank. If False, it uses the true (tree sequence)
     times. If None, times from the tree sequence are not used and the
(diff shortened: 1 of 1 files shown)


argutils/viz.py
```diff
@@ -30,25 +34,28 @@ def draw(ts, ax, use_ranked_times=None, tweak_x=None, arrows=False):
         except (TypeError, KeyError):
             labels[nd.id] = str(nd.id)
 
-    pos = nx_get_dot_pos(G)
+    positions = nx_get_dot_pos(G)
```
jeromekelleher, line 37: This would be a lot simplier/easier to follow if it was either/or. Why do you want to allow updating pos dictionary?

jeromekelleher (commented):
Good idea, but I don't understand the need for the complicated semantics?

argutils/viz.py
hyanwong: I assumed that we might want to only change some of the node positions (although I guess normally we would change all of them). It seemed easy enough to simply replace all the existing positions, but actually we could do that with positions.update(pos) rather than an explicit loop.

jeromekelleher: But we're not *actually* using it, right? These PRs would be a lot quicker if they included only functionality that we need. 

I know this is annoying, but it's far easier for me or anyone else to understand these functions if they just have the functionality that's being used. Adding on loads of bells and whistles just because we can and we might want them later makes it much harder to follow. 

It also makes them much harder to change. Suppose I wanted to add some functionality that conflicted with this - how would I know that this is a "we might want this later" feature, which can be dropped vs something that's actively being used?

jeromekelleher: If we do want it later, it's a 30 second job to do it. It's not like we've written test cases to cover the functionality or anything, so it really is just a waste of time putting in extra stuff before it's needed.

hyanwong: Sure: this just seemed like the most obvious way to do it when it occurred to me.

hyanwong: Done

Issue #82, opened 2022-02-09
Figure 5: Which nodes are optional in an edge-annotated arg


hyanwong:
Here's what we have at the moment
<img width="566" alt="Screenshot 2022-02-09 at 14 50 09" src="https://user-images.githubusercontent.com/4699014/153225761-6a8b4ade-4cb1-4e5e-923b-034ea0e282ca.png">

but we want to replace this with an arg simulated using the `argutils.sim_coalescent` method. I suggest using the one below (from [here](https://github.com/tskit-dev/what-is-an-arg-paper/issues/80)). We need to think about what node colour we use, including representing the length over which a node is coalescent as a shade of colour. We should also think what we say about the pass-through nodes (these are mostly the 2 parents that exist for each recombinant node). In the plot above we get rid of these by showing a MultiDiGraph, where e.g. a diamond can be shown by two edges joining a parent to a child node.

```
import argutils
import matplotlib.pyplot as plt
import tskit
ts = argutils.viz.label_nodes(argutils.sim_coalescent(4, 0.001, 1000, seed=123))
print(f"{ts.num_samples} samples, {ts.num_trees} trees")
fig, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(20,10))
ax1.set_title("(a) full coalescent arg")
ax2.set_title("(b) simplified keeping unary in coal")
ax3.set_title("(c) fully simplified")
pos = argutils.viz.draw(ts, ax1, use_ranked_times=True, tweak_x={24: -15, 23: -5, 22: -10, 21: -10})
new_ts, node_map = simplify_keeping_unary_in_coal(ts, map_nodes=True)
argutils.viz.draw(new_ts, ax2, pos={node_map[i]: p for i, p in pos.items()})
new_ts, node_map = ts.simplify(map_nodes=True)
argutils.viz.draw(new_ts, ax3, pos={node_map[i]: p for i, p in pos.items()})
```

![image](https://user-images.githubusercontent.com/4699014/153241204-1fdc672b-2566-4f13-ad54-3c5023004fb7.png)<img width="736" alt="Screenshot 2022-02-09 at 15 05 38" src="https://user-images.githubusercontent.com/4699014/153228679-50b14c86-f716-49d3-bc04-77b51e7c7661.png">

[...]

PR #85, opened 2022-02-09
draw edge spans as graph edge widths

awohns:
One improvement to the graph visualizations that was discussed with @jeromekelleher and @hyanwong was to draw the width of edges in the graph representation of the tree sequence proportional to the genomic span of edges. This PR has a simple update to the `viz.py` to support this.

Here's @hyanwong's example from #82 updated with edge spans.

```
ts = argutils.viz.label_nodes(argutils.sim_coalescent(4, 0.001, 1000, seed=123))
fig, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(20,10))
ax1.set_title("full coalescent arg")
ax2.set_title("simplified keeping unary in coal")
ax3.set_title("fully simplified")
pos = argutils.viz.draw(ts, ax1, use_ranked_times=True, tweak_x={24: -15, 23: -5, 22: -10, 21: -10}, draw_edge_widths=True)
new_ts, node_map = argutils.simplify_keeping_unary_in_coal(ts, map_nodes=True)
argutils.viz.draw(new_ts, ax2, pos={node_map[i]: p for i, p in pos.items()}, draw_edge_widths=True)
new_ts, node_map = ts.simplify(map_nodes=True)
argutils.viz.draw(new_ts, ax3, pos={node_map[i]: p for i, p in pos.items()}, draw_edge_widths=True)
```
Screen Shot 2022-02-09 at 1 05 03 PM

I think this gives good intuition about the shape of the tree sequence: especially evident with a larger simulation:
Screen Shot 2022-02-09 at 1 06 11 PM


awohns:
@hyanwong mentioned the other option (only requiring a slight modification of this PR) of drawing edge colors in grayscale proportional to their span. This is what the above two figures would look like in grayscale:

Screen Shot 2022-02-09 at 1 27 53 PM
Screen Shot 2022-02-09 at 1 28 20 PM

awohns:
@hyanwong here's the grayscale version again with thicker edges:
Screen Shot 2022-02-09 at 2 05 13 PM
Screen Shot 2022-02-09 at 2 05 39 PM

jeromekelleher:
This is fabulous @awohns, we're really starting to see some structure coming through here! It's also clear that there are a lot of advantages to the "keep unary in coal" version (which is more-or-less what tsinfer outputs, I think).

My favourite is the first version. The line-thickness is a very clear indication of what's going on, and it's the most robust to viewing on screen vs paper etc.

argutils/viz.py
```diff
@@ -144,3 +155,18 @@ def nx_get_dot_pos(G):
         if n.get_pos() and n.get_name().isdigit()
     }
 
+def get_edge_spans(G, ts, max_edge_width):
+    """
+    Returns a list of length "num_edges" in G, the graphical representation of the
+    tree sequence, where each entry is the genomic span of the corresponding edge in the
+    tree sequence.
+    """
+    edge_spans = (ts.tables.edges.right - ts.tables.edges.left)
```
jeromekelleher, line 164: This isn't quite right as we're thinking about graph theoretic edges here (u, v) rather than tskit ones (left, right, u, v).

The nx graph should already have an attribute ``intervals`` which has all (left, right) pairs associated with it. See the convert_nx (or whatever it's called) for details. You should be able to easily compute from that.

jeromekelleher: There's a discussion about whether it should be a multigraph for viz purposes, but let's not worry about that for now.

jeromekelleher (commented):
Looks great @awohns! I think there's a slight logic bug, though.

hyanwong:
Looks great Wilder, thanks. I like the third version best, but I could be talks round. I think we want the viz to work even with tiny edges, and if there are multiple edges between close nodes, so I'm a little worried about the ones that have very thick lines.

I could be convinced either way though. Let's see what other people think?

hyanwong:
There's an interesting viz point here, which is that if we can plot out the *implicit* tree sequence, we would have thick black lines everywhere (apart from those from recombination nodes). This might be a nice way to illustrate the conversion from implicit to explicit via simplification? But after https://github.com/tskit-dev/what-is-an-arg-paper/pull/73 I think we can't output an implicit ARG from the argils.sim_xxx routines, right?

jeromekelleher:
sim_coalescent has a "resolved" option for just this. You can't do with with the WF simulator though (wouldn't be hard to do, if we needed it, though).

hyanwong:
> sim_coalescent has a "resolved" option for just this

Oh, perfect. Apologies, I misread that that PR had removed it from the `sim_coalescent` function too.

awohns:
I decided to include both options: drawing the edge widths proportional to edge span and/or to draw the alphas proportional to spans, so we can decide later what looks better in a figure.

Also, thanks for catching the logic bug @jeromekelleher. I've fixed that as well.

I think this is ready to merge

argutils/viz.py
```diff
@@ -174,3 +205,23 @@ def nx_get_dot_pos(G, reverse_x_axis=None):
...
+def get_edge_spans(G):
+    """
+    Returns a list of length "num_edges" in G, the graphical representation of the
+    tree sequence, where each entry is the genomic span of the corresponding edge in the
+    tree sequence divided by the total sequence length.
+    """
+    edge_spans = list()
+    for edge in G.edges():
+        edge_data = G.get_edge_data(edge[0], edge[1])
+        edge_spans.append(edge_data["intervals"][0][1] - edge_data["intervals"][0][0])
```
jeromekelleher, line 218: This is assuming you've only one interval per edge though. Should be something like
```
total_span = sum(right - left for (left, right) in edge_data["intervals"])
```

argutils/viz.py
```diff
@@ -76,6 +88,19 @@ def draw(
...
+
+    if draw_edge_widths:
+        edge_widths = get_edge_spans(G)
+        sequence_length = ts.get_sequence_length()
+        edge_widths = (edge_widths / sequence_length) * max_edge_width
+    else:
+        edge_widths = None
+    if draw_edge_alpha:
+        edge_alpha = get_edge_alpha(G, ts)
+        if draw_edge_widths is False:
```
jeromekelleher, line 100: ``not draw_edge_widths``

jeromekelleher: Perhaps more logical to use the ``max_edge_width`` here rather than "3"?

jeromekelleher (commented):
Looks good. Still a logic bug how edge widths are computed, but ready to merge otherwise.

hyanwong:
It would be great if you could sort & merge this @awohns - I think it's almost there, right?

awohns:
Incorporated @jeromekelleher's changes, I think it's good to go now. I don't have write access so can't merge the PR.
Thanks both!

PR #88, opened 2022-02-10
Add a ``recomb_proba`` arg to wright fisher sim.

### LINKED ISSUE #87: WF simulator should be able to set recombination rate
As discussed in [#82 ](https://github.com/tskit-dev/what-is-an-arg-paper/issues/82#issuecomment-1034121083), otherwise we end up simulate an entire chromosome by default, which results in ARGs which are hard to visualise.

I'm happy to do this, but OTOH it might be a 2 min job for @jeromekelleher. Just opening this as an issue so that we don't forget. Simple to close when done.

jeromekelleher:
Closes #87

It didn't seem worth the effort of thinking through what a proper rate would be here, so I just chucked in a simple probability.

### DIFF
--- a/argutils/ancestry.py
+++ b/argutils/ancestry.py
@@ -320,9 +320,17 @@ def simplify_keeping_unary_in_coal(ts, map_nodes=False):
         return tables.tree_sequence()
 
 
-def sim_wright_fisher(n, N, L, seed=None):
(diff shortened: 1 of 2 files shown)


jeromekelleher:
Merging - we can always add something better if we want it.

PR #89, opened 2022-02-11
Draft of figure 2

jeromekelleher:
Also some tweaks to the text.

The figure draft is *very* rough, and really just getting in the basic information I want to discuss in the section. I think it's worth getting it in now, and writing the text and iterating before trying to make it look any better.

argds


hyanwong:
Looks good, but it would also be good to remove the infinity signs, I think?

hyanwong:
I very much like the way the caption is worded, BTW.

jeromekelleher:
But lets see how that comes out in the text and we can revise if it's not clear

hyanwong:
> I want to convey the idea that you have to implement the distinction between the different types of nodes somehow - breakpoint(u) = infinity is one way of saying "u is not a recombination node".

"NA" or simply "-" would probably be more obvious to the casual reader?

jeromekelleher:
Probably - let's not worry about it for now.

hyanwong:
I wonder if we somehow want to point out visually that the nodes on the LH diagram are of a different type to the nodes on the right. I guess we could use letters, but that's means comparisons are difficult. Perhaps bold vs non bold, or italic, or sans vs sans serif?

jeromekelleher:
Maybe - but I would rather do some writing now before trying to improve it.

hyanwong:
> Maybe - but I would rather do some writing now before trying to improve it.

True. Why not merge and I'll open up an issue where I can put in a few suggestions so they aren't lost.

jeromekelleher:
> True. Why not merge and I'll open up an issue where I can put in a few suggestions so they aren't lost.

That would be great,thanks

PR #91, opened 2022-02-11
Fix error in WF sim

jeromekelleher:
Fix for the problem in the WF sims you pointed out earlier @hyanwong


hyanwong:
Thanks a lot @jeromekelleher

PR #93, opened 2022-02-11
working through ARG representation section

jeromekelleher:
Rough pass on this section which is good enough as a first draft. Rather than fixating on the detail here, I think it would be good to  plough on with drafting other sections and then circle back when we have a good picture of the overall narrative and how we want to structure it.


jeromekelleher:
Going to merge this in case anyone is inspired to work on the text over the weekend!

PR #94, opened 2022-02-11
Implement 3 new types of simplification

hyanwong:
Implement `test_simplify_keeping_all_nodes`, `simplify_remove_pass_through` (with option to repeatedly apply until no pass through nodes remain), and tidy up test_simplify_keeping_unary_in_coal so that it runs OK on tree sequences with individuals already present (e.g. for WF simulations).

The `simplify_remove_pass_through` function turned out to be a bit tricky. I think I have it right though, and there are tests for all this.

### DIFF
--- a/argutils/tests.py
+++ b/argutils/tests.py
@@ -1,4 +1,6 @@
 # Run using python3 -m pytest argutils/tests.py
+import collections
+import string
 
 import tskit
 import pytest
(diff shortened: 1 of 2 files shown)


hyanwong:
Merging this so I can PR a version of Fig 5 with the various simplification options.

argutils/tests.py
```diff
@@ -131,3 +133,97 @@ def test_examples(self, unresolved, resolved):
...
+        for e in ts3.edges():
+            node_edges[e.child][0] += 1
+            node_edges[e.parent][1] += 1
+        for i, row in enumerate(node_edges):
+            # No more pass thoguh nodes left
+            assert list(row) != [1, 1]
+        
+        
+    def test_simplify_keeping_unary_in_coal(self):
+        for ts in [
```
jeromekelleher, line 207: Just FYI: pytest.mark.parametrize is much better for this, and can follow exactly the same pattern.

hyanwong: Oh yes, I forgot about that. Thanks! I'll make a follow up PR.

PR #95, opened 2022-02-11
One node recombination demo

hyanwong:
So I wanted to see if I could make a version of the Wiuf and Hein ARG with only one node at the recombination breakpoint, rather than 2, to test out @jeromekelleher 's point about 2 nodes being required for simulation backwards-in-time (https://github.com/tskit-dev/what-is-an-arg-paper/discussions/41). I'm *not* suggesting that this is how we should simulate stuff: this PR is simply to kick the wheels of the various ideas we've been playing with.

What I've put together is just a demo, and need not be merged, but I think it reveals an interesting point. If you have a node annotated ARG, then I think you might be able to get away with simulating backwards in time using a single breakpoint, because you limit the span of the edge depending on whether the *child* node has a breakpoint (and you look up that breakpoint position when you add the edge). This produces a different sort of "implicit arg", where the edges *above* the recombination event are from 0..x and x..L. In the 2-node version, the edges *below* the recombination event are the ones that are restricted like this. I suspect that the two different encodings will give identical simplified tree sequences.

I may well be missing something in how the coalescent simulation works, since this is a hard-coded example, but I would be interested to know if the one-node version could be implemented in the `argutils.sim_coalescent` code. The weird thing would be, I think, that you would decide (arbitrarily, I guess) whether to transmit the left or right portion of the genome  *when you added the edge from the parents back to the child (recombination) node*, not when you generate the recombination event itself. The next time you pick that node as the child of a different parent, you would have to, I suppose, take the other portion, so there would be some nasty state storing on the node required. Nevertheless, it is in principle doable (I think?)

FWIW, here's some demo plots using `argutils.wh99_example(one_node_recombination=True)`:

```
import argutils
import tskit
import numpy as np
from IPython.display import SVG

fig, (ax1, ax2, ax3, ax4, ax5) = plt.subplots(1, 5, figsize=(20,10), sharey=True)

ax1.set_title("(a) big ARG with implicit\nencoding (Wiuf & Hein)")
ts = argutils.viz.label_nodes(argutils.wh99_example(one_node_recombination=True))
pos = argutils.viz.draw(ts, ax1, use_ranked_times=True,
                        #tweak_x={0: 5, 3: -20, 4:-10, 2: -10, 6: 15, 9: 10, 14: 15}
                       )

ax2.set_title("(b) explicit encoding\n(i.e. non-ancestral removed)")
ts2 = argutils.simplify_keeping_all_nodes(ts)
argutils.viz.draw(ts2, ax2, pos=pos)

ax3.set_title("(c) pass-through nodes\nremoved")
ts3, node_map = argutils.simplify_remove_pass_through(ts2, repeat=True, map_nodes=True)
argutils.viz.draw(ts3, ax3, pos={node_map[i]: p for i, p in pos.items()})

ax4.set_title("(d) nodes removed if\nnon coalescent everywhere")
ts4, node_map = argutils.simplify_keeping_unary_in_coal(ts, map_nodes=True)
argutils.viz.draw(ts4, ax4, pos={node_map[i]: p for i, p in pos.items()})

ax5.set_title("(e) full simplification")
ts5, node_map = ts.simplify(map_nodes=True)
argutils.viz.draw(ts5, ax5, pos={node_map[i]: p for i, p in pos.items()})
```

image

(compare with `one_node_recombination=False`:)

image


hyanwong:
NB: I think that this way of doing it is very much dependent on node-annotated breakpoints, so is not really how you would want to do it in an edge-annotated world. But it's certainly one potential approach that other people might take / have taken.

hyanwong:
I bring it up because *visually* I think our version needs considerable explanation to those coming from the ARG with events-as-nodes world. For example, if I were to look at the figure in Wiuf and Hein, I would find it difficult to translate to the figure 5 we have at the moment. And I would be asking why one of the nodes (J) disappears on simplification, given that this is a Hudson ARG, so is not expected to have redundant nodes (the equivalent recombination node doesn't not disappear on simplification of the 1-RE-node version, of course).

I'm just about to post on slack about something related to this though, so maybe hold off.

jeromekelleher:
> I bring it up because visually I think our version needs considerable explanation to those coming from the ARG with events-as-nodes world. 

We need to have resolved that issue by the time we get to this point in the manuscript. That's why we're starting off with nice familiar diagrams, and showing how we gain much more expressivity and power if we make two small changes. We can assume that people have been convinced at this point that people have two parents, rather than one.

I think it would be far worse to compromise what we ultimately believe is the right way to do things just to make something look a bit more familiar.

hyanwong:
There's also the slightly odd thing about node J and its ilk disappearing from the (Hudson) graph. I guess we could either skate over that, or use it as an explanatory point about ignoring nodes on lineages with no ancestral material

jeromekelleher:
Reposting this @hyanwong 

>If you want one single reason for why single node recombination is bad, consider a simple diamond. If we have single node recombination, then the breakpoint will be lost unless we consider squashable edges significant. I absolutely don't want this to be the case, because we current don't, and it would greatly complicate msprime and algorithms like simplify if we did. It also means that we can't use set theoretic operations when thinking about the intervals associated with edges, because (x, y] union (y, z] == (x, z] under any sensible definitions of those things.

This has nothing to do with simulation.

hyanwong:
> This has nothing to do with simulation.

Yes, that's true. Very good point.

jeromekelleher:
I think we can probably close this @hyanwong?

hyanwong:
Yep

hyanwong:
I'm going to merge this now, as slack discussion has approved of the general approach

Issue #36, opened 2022-01-19
Figure showing relationship between ARG and pedigree


jeromekelleher:
From discussions in #22 and offline with @hyanwong, it seems clear that our first figure should be one showing the relationship between the ARG and the pedigree (like Brandt et al, and Mathieson and Scally).

We decided that a good structure to aim for would be three panels, ideally generated from a simulated example:

1. The pedigree relating individuals
2. The *genetic* pedigree showing the embedded graph of inheritance between genomes
3. The ARG, which is a subset of the genetic pedigree (pass through nodes are removed) and with recombination breakpoints added as annotations on the appropriate nodes.

How does this sound @hyanwong - is it feasible to do with networkx and simulations? We could just copy a simple WF pedigree simulator from msprime (this one? https://github.com/tskit-dev/msprime/blob/df31bc05dea7bba516ef2dd073cc47a00c827ca5/msprime/pedigrees.py#L103) rather than adding in complex dependencies.

jeromekelleher:
I guess the drawback here is that the msprime fixed_pedigree model doesn't support recording the ARG nodes. We could hack the python version to do so pretty easily, and copy it into this repo? (Or just make the change if it's not too hard - I suspect it'll be time consuming to do properly though, as we'll have to test it properly and consider the interactions with the other options.)

hyanwong:
We probably wouldn't simulate it in msprime anyway. I'd use SLiM. That way we can record all the individuals (including the non-genetic ones), which is useful when showing the pedigree, but have the embedded tree sequence with unary nodes too, so we don't need to rely on simulating via msprime through the pedigree?

jeromekelleher:
OK, whatever's simplest.

jeromekelleher:
From today's discussion, this is Fig 3, and @hyanwong has agreed to pick it up.

We decided to show a (very) inbred pedigree with numbered node IDs in each of the individuals. It's best to do this as a hand-drawn figure.

This is a bit different to what's discussed above.

Sound good @hyanwong?

hyanwong:
We also decided (I think) to use letters for the node labels, although that can easily be switched.

hyanwong:
Can you post the picture you took of the whiteboard in this issue, please @jeromekelleher ? (And I guess the appropriate photos in the other issues too)

jeromekelleher:
[...]

PR #97, opened 2022-02-13
Add figure 3

hyanwong:
Draft for https://github.com/tskit-dev/what-is-an-arg-paper/issues/36


hyanwong:
Waiting on #98 for a better graph plot

illustrations/pedigree.py
```diff
@@ -0,0 +1,76 @@
+import json
+from pathlib import Path
+import string
+import sys
+from types import SimpleNamespace
+
+import matplotlib.pyplot as plt
+import tskit
+
+# useful modules in the top level dir
```
jeromekelleher, line 10: You don't need to do this - just run ``python illustrations/pedigree.py`` from the top level dir,

jeromekelleher (commented):
This isn't compiling for me - there's no rule to build ``pedigree.pdf``.

Life would be easier if we had a single file ``illustrations.py`` that lived in the top level directory and that produced the different illustrations via subcommands (using click).  Having code mixed in with the output files in ``illustrations`` is making things confusing I think.

illustrations/pedigree.py
hyanwong: right, as long as we remember that's where it's run from, I guess. On the command-line I find it easier to run the things separately from within illustrations.

jeromekelleher: Let's move all the code for illustrations into one file, ``illustrations.py`` as suggested. Everything gets run from the root, always. Nothing to get confused about then.

hyanwong:
> There should be a pedigree.svg in `illustrations`?

Forgot to git add it, that's all. Sorry. Updated

jeromekelleher:
Works for me now. Look great, thanks. Merge away whenever you're happy.

illustrations/pedigree.py
hyanwong: Oh yes, that's why I did it like this: it's useful to be able to copy and paste the code into a jupyter notebook, as I template and re-test most of the figure code in notebooks which is a lot more responsive than using a script.

I guess I could simply dump the code into scripts and call them within the `illlustrations.py` script. Unless you have better suggestions for easy conversion between notebook cells and free-standing scripts?

jeromekelleher: Why can't you do that here? You can copy a function definition into a notebook, can't you?

hyanwong: Oh yes, I guess I could simply turn it into a function. Good point.

PR #99, opened 2022-02-13
Tweaks to Fig 1

### LINKED ISSUE #96: Tweaks for Fig 1
Currently Figure 1 looks like this:

arg-process

- I think it would be better if this used the language of "events" and avoided drawing two recombination nodes, since we're not broaching the idea of using nodes-as-genomes until Figure 2. We could remove all the "dots" and only label a few key events that we are interested in discussing? (This would make the diagram look more like the classical ones, too)
- I hacked the layout to make it go side-by-side - it would be good to make it a little less hacky, if possible? (Ideally all in one tikz figure, but that's possibly too awful to contemplate)
- "Extant material" is maybe not the best. What if we did two columns for the two competing rates, headed "CA" and "RE" (maybe)? E.g. $\binom(4)(2)$ and $2L\rho$?

What do you think @JereKoskela ?

JereKoskela:
Removed nodes from Figure 1 and tweaked the rate columns. I haven't changed the side-by-side hack because while putting all of this into one tikz diagram is probably doable, the result will be impossible to tweak further. 

Closes #96

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -142,101 +142,79 @@ description of the algorithm, and how it may be implemented efficiently.
 \scalebox{0.5}{
 \begin{tabular}{cc}
 \begin{tikzpicture}
-	% Samples at x = 0, 4, and 6
-	\node [circle, fill, scale=0.2,label=above right:{\small A}] at (0,0) {A};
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Great, thanks.

PR #103, opened 2022-02-13
Mention "big" vs "little" ARG earlier

hyanwong:
I'm not sure if this works or not, but I think it is much clearer if we switch to "big ARG" vs "little ARG" terminology as soon as possible, because we bandy the word "ARG" around when talking about Griffiths stuff, and it gets pretty confusing (to me, at least).

This is my attempt to try that out, but I'm not sure if it seems too clunky or not.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -115,6 +115,9 @@ edges rather than nodes.
 \item Suggest a classification of ARG nodes that helps us to understand
 the properties of the structures inferred by different methods, and illustrate
 with some examples.
+% Perhaps we should also mention something about simplification and unknowable
+% nodes, e.g. "show how some of the nodes are not required for useful interpretation
+% of ancestry".
 \end{enumerate}
 
 \section*{Ancestral graphs and stochastic processes}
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -433,17 +436,24 @@ \section*{Ancestral graphs and stochastic processes}
 closely related and complementary treatment of the same process. Where
 Hudson's algorithm is focused on the efficient \emph{simulation} of the
 process, the original ARG literature is focused on mathematical
-results about the stochastic process. Griffiths and colleagues formulated the
-ARG as a branching-coalescing stochastic process. We imagine a graph in which
-each edge corresponds to an extant lineage and nodes are events in the
-process (the initial $n$ leaf nodes are ``sampling'' events). As in Hudson's
+results about the stochastic process. Griffiths and colleagues used the word
+"ARG" to refer to both a specific branching-coalescing stochastic process and
```
jeromekelleher, line 440: Can you use TeX quotes please, ``like this''.

jeromekelleher (commented):
I don't think this really captures it to be honest. The Big ARG can only really refer to the original stochastic process defined by Griffiths. In practical terms, there's no such thing as a Big ARG which contains nodes with no ancestry - nobody simulates this and certainly nobody infers it.

I actually don't know where the terms Big and Little ARG came from. Maybe it's just a convention people have adopted in person. Maybe it would be better if we didn't use them at all, to be honest.

So, the original ARG-as-a-stochastic process == Big ARG, and we don't bother mentioning Little ARGs.

We will have to be able to use the term ARG to mean different things in different sections of the paper though, this is just inevitable. We probably should be up front about that.

So, I don't think is it, but we can merge this much if you like as a step in the right direction.

paper.tex
hyanwong: I just replaced those, and see that a number of other places use double quotes to end ``like this" rather than ``like this''. Do you want me to replace those too?

hyanwong:
> I actually don't know where the terms Big and Little ARG came from. Maybe it's just a convention people have adopted in person. Maybe it would be better if we didn't use them at all, to be honest.

I'm not too bothered about *what* terminology we use, but I think we do need to establish some qualifying adjectives very early on to prefix onto the word "ARG", otherwise it gets very confusing. Perhaps that could be the "Griffiths ARG", or something else, perhaps the "Griffiths process" vs the "Griffiths graph"?. A phrase like "The ARG is vast, and any realisation for even moderate levels of recombination is far too large to be of practical" needs some sort of qualification, because we are definitely talking about the Griffiths one here, and not HUD, or a tree sequence representation of a recombination graph. 

Also, I thought that we wanted to argue that we should reserve the word ARG for a graph, rather than the process. So we want to be rather careful after the first mention on line 433 I think. As you say, put the [definitions](https://github.com/tskit-dev/what-is-an-arg-paper/discussions/23) as up front as possible. And in those definitions, we use ARG for the graph, not the process.

> So, the original ARG-as-a-stochastic process == Big ARG, and we don't bother mentioning Little ARGs.

Fine, but I think we need a word for HUD don't we? And I don't like HUD much. I think "little ARG" is fine TBH, but I'm quite happy with other suggestions. I think one of the points here is that we are claiming that a tree sequence is a type of ARG (right?), so we want to move away from the word "ARG" without any qualifiers ASAP.

> So, I don't think is it, but we can merge this much if you like as a step in the right direction.

Yes, I agree that I don't; quite have it right here. I'm happy to merge and iterate, or thrash out a better phrasing below. 

Perhaps we should refer to

jeromekelleher:
Happy to merge. I'm going to focus on getting all the content into a first draft before getting caught up in the details - sorting out the terminology is a second-pass operation I think.

Issue #56, opened 2022-02-02
Figure 4: genome coordinate annotations


jeromekelleher:
We want a figure showing how the implicit genome coordinate annotations that we use in our ARG encoding (each recombination event with child u and breakpoint x has two parents, v and w, and we record edges (-inf, x, v, u), (x, inf, w, u)) is resolved to fully explicit edges via a variant of Hudson's algorithm. 

Panel A is an ARG encoded in the form above, and panel B is the fully resolved equivalent.

@a-ignatieva - are you happy picking this one up?

a-ignatieva:
Yep, leave it with me!

a-ignatieva:
How's something like this? Left is implicit coordinates, boxes hopefully show that the breakpoints are "local". Right is explicit coordinates, green is bits of the genome that have reached mrca.
![drawing](https://user-images.githubusercontent.com/36134434/152687482-7dcc64ee-b72e-47f3-8b07-83f7bb995040.svg)

hyanwong:
Nice. I like it. It would be good to have some terminology to distinguish between these two forms (did we decide on implicit vs explicit?). See https://github.com/tskit-dev/what-is-an-arg-paper/discussions/23#discussioncomment-2120250

jeromekelleher:
This is really excellent, thanks @a-ignatieva! There's a few things we need to sort out terminology and notation-wise (mainly the infinities), so let's not finalise it yet. I'll get back with some concrete suggestions once we've figured that out.

Another thing to consider is that what we're producing is very like Fig 1 in "Recombination as a point process", which we should acknowledge. I'll reproduce here for convenience:

![HUD](https://user-images.githubusercontent.com/2664569/152767158-6e387f6e-454c-4558-9d94-0128d121c318.png)

jeromekelleher:
We could just reuse the example and recast it in our new notation? It's a great example!

jeromekelleher:
See #64 for code to produce the wh99 ARG.

jeromekelleher:
That's true - let's hold off a bit until we've figured out the inf-vs-0/L issue.

hyanwong:
To pick up a point in https://github.com/tskit-dev/what-is-an-arg-paper/pull/85#issuecomment-1036066898, we could possibly use line width or intensity (alpha values?) to show the span of the edges?

We also need to revise this to use `0/L` rather than `inf`. I'm happy to do that if @a-ignatieva is busy

[...]

PR #109, opened 2022-02-14
First stab at ancestry resolution text

jeromekelleher:
WIP

### DIFF
--- a/argutils/ancestry.py
+++ b/argutils/ancestry.py
@@ -12,8 +12,6 @@ import numpy as np
 import tskit
 import networkx as nx
 
-from . import viz
-
 NODE_IS_RECOMB = 1 << 1
 
 
--- a/argutils/ancestry.py
+++ b/argutils/ancestry.py
@@ -8,6 +8,7 @@ import dataclasses
 from typing import List
 from typing import Any
 
+import intervaltree
 import numpy as np
 import tskit
 
@@ -577,6 +578,65 @@ def wh99_example():
     return tables.tree_sequence()
 
 
+class IntervalSet:
+
+    intervals: list = dataclasses.field(default_factory=list)
+
+    def contains(self, x):
+        for left, right in self.intervals:
+            if left <= x < right:
+                return True
+        return False
+
+    def union(self, other):
+        """
+        Returns a new IntervalSet with the union of intervals in this and
+        other.
+        """
+        # Dreadful implementation, but oh well.
+        intervals = self.intervals.copy()
+        for interval in other.intervals:
+            if interval not in intervals:
+                intervals.append(interval)
+        return IntervalSet(intervals)
+
+
+import intervaltree
+
+def as_garg(ts):
+    """
+    Returns the specified unresolved ARG as an GARG E
+    """
+    E = [(edge.child, edge.parent, intervaltree.IntervalTree.from_tuples([(edge.left, edge.right)]))
+        for edge in ts.edges()]
+    return E
+
+def as_resolved_garg(ts):
+    """
+    Returns the specified unresolved ARG as an GARG E with respect to
+    the tree sequences samples.
+    """
+    S = ts.samples()
+    E = as_garg(ts)
+    print(E)
+    I = collections.defaultdict(intervaltree.IntervalTree)
+    for u in S:
+        I[u] = intervaltree.IntervalTree.from_tuples([(0, ts.sequence_length)])
+
+    N = set(S)
+    while len(N) > 0:
+        c = N.pop()
+        for e in [e for e in E if e[0] == c]:
+            p = e[1]
+            print("parent = ", p)
+            print("Ip = ", I[p])
+            print("Ic = ", I[c])
+            print("e = ", e[2])
+            print("intersection = ", I[c].intersection(e[2]))
+            I[p] = I[p].union(I[c].intersection(e[2]))
+            N.add(p)
+    return [(c, p, I[c]) for c, p, _ in E]
+
 def as_earg(ts):
     """
     Returns the specified unresolved ARG as an EARG (E, sigma).
@@ -621,4 +681,22 @@ def earg_get_tree(E, sigma, S, x):
                 p = P[1]
             parent[c] = p
             N.add(p)
+    return parent
+
+
+def garg_get_tree(E, S, x):
+    """
+    Given a minimal GARG definition E, return the tree for a the given
+    set of samples at the specified position as a dictionary parent->child.
+    """
+    N = set(S)
+    parent = {}
+    while len(N) > 0:
+        c = N.pop()
+        P = [e[1] for e in E if e[0] == c and e[2].overlaps(x)]
+        # print(c, P)
+        if len(P) > 0:
+            assert len(P) == 1
+            parent[c] = P[0]
+            N.add(P[0])
     return parent
--- a/argutils/ancestry.py
+++ b/argutils/ancestry.py
@@ -10,6 +10,7 @@ from typing import Any
 
 import numpy as np
 import tskit
+import networkx as nx
 
 from . import viz
 
@@ -653,20 +654,17 @@ def as_resolved_garg(ts):
     Returns the specified unresolved ARG as an GARG E with respect to
     the tree sequences samples.
     """
-    G = viz.convert_nx(ts)
-    import networkx as nx
-
-    topological = list(nx.topological_sort(G))
-    # print(topological)
-    Q = collections.deque(ts.samples())
     E = as_garg(ts)
+    G = nx.DiGraph()
+    G.add_edges_from([e[:2] for e in E])
+    # We have to get the topological sorting of the nodes because we don't
+    # have time ordering.
+    topological = list(nx.topological_sort(G))
+
     I = collections.defaultdict(lambda: IntervalSet(ts.sequence_length))
-    for u in Q:
+    for u in ts.samples():
         I[u] = IntervalSet(ts.sequence_length, [(0, ts.sequence_length)])
 
-    # while len(Q) > 0:
-    # print("Q = ", Q)
-    # c = Q.pop()
     Ep = []
     for c in topological:
         # print("c = ", c)
--- a/argutils/ancestry.py
+++ b/argutils/ancestry.py
@@ -575,3 +575,50 @@ def wh99_example():
 
     tables.sort()
     return tables.tree_sequence()
+
+
+def as_earg(ts):
+    """
+    Returns the specified unresolved ARG as an EARG (E, sigma).
+    """
+    sigma = np.full(ts.num_nodes, int(ts.sequence_length), dtype=int)
+    edges = iter(ts.edges())
+    edge = next(edges, None)
+    E = []
+    while edge is not None:
+        if edge.left == 0 and edge.right == ts.sequence_length:
+            E.append((edge.child, edge.parent))
+        else:
+            assert edge.left == 0
+            breakpoint = edge.right
+            child = edge.child
+            sigma[child] = breakpoint
+            E.append((edge.child, edge.parent))
+            edge = next(edges, None)
+            assert edge.left == breakpoint
+            assert edge.child == child
+            E.append((edge.child, edge.parent))
+        edge = next(edges, None)
+    # We could wrap sigma as a function to be literal about the definition
+    # but this is simpler for debugging.
+    return E, sigma
+
+
+def earg_get_tree(E, sigma, S, x):
+    """
+    Given a minimal EARG definition (E, sigma), return the tree for a the given
+    set of samples at the specified position as a dictionary parent->child.
+    """
+    N = set(S)
+    parent = {}
+    while len(N) > 0:
+        c = N.pop()
+        P = [e[1] for e in E if e[0] == c]
+        if len(P) > 0:
+            if x < sigma[c]:
+                p = P[0]
+            else:
+                p = P[1]
+            parent[c] = p
+            N.add(p)
+    return parent
--- a/argutils/ancestry.py
+++ b/argutils/ancestry.py
@@ -8,9 +8,10 @@ import dataclasses
 from typing import List
 from typing import Any
 
-import intervaltree
 import numpy as np
 import tskit
+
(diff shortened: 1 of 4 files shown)


jeromekelleher:
Here's a stab at some formal definitions (brought togther in one place for clarity here):

formal-defs

Any thoughts here @JereKoskela @hyanwong? I'm trying hard for mathematical economy here, no necessarily mapping to computational implementations.

jeromekelleher:
Obviously would have to define what we mean by "through the local trees" here.

hyanwong:
Rather than using N (natural numbers), do we want to define N to be the set of all nodes? Then it's a bt more obvious to say S is a subset of the set of all nodes.

Also, there is a constraint on the eARG that says that there only two parents per child. In the original formulation, I suspect there is also the constraint that there are only 2 children per parent, although that's not strictly necessary.

hyanwong:
> Obviously would have to define what we mean by "through the local trees" here.

Do we need to say "a path *down* from c" (and do we need to indicate that this is a DAG, i.e. that the order of (c, p) matters)

> The "non-empty subset" is also intended to imply that c is not a root

Are we not also allowing nodes above the root? I think that';s OK, right? Just not what we normally bother doing.

hyanwong:
Presumably we also define what an implicit vs an explicit gARG is?

jeromekelleher:
> But rather than using N (natural numbers), do we want to define N to be the set of all nodes? Then it's a bt more obvious to say S is a subset of the set of all nodes.

I don't think we need to define the set of nodes, it's implicitly defined by whether a number exists as a child or parent in E. We probably don't need to say that S is a subset of the natural numbers though, and can just say "S is a set of nodes".

Good point on the restrictions on the EARG, they need to be stated.

jeromekelleher:
> Presumably we also define what an implicit vs an explicit gARG is?

That's the distinction between a GARG and a sample resolved GARG. You can express a GARG without defining any samples. Then all you have is the local parent-child relationships. If you define a set of samples, you can then resolve the ancestry wrt to that set.

jeromekelleher:
> Are we not also allowing nodes above the root(s)? I think that's OK, right? Just not what we normally bother doing.

Ah, you're right, that's full simplification I'm thinking about. No need for that then.

jeromekelleher:
> It's the same tuple as in the GARG case, subject to a further constraint, right? In which case we could economise a bit and just say that.

Yes, strictly it's the same exception the interval I_{resolved} will be \subseteq I_{not resolved}.

hyanwong:
> > Presumably we also define what an implicit vs an explicit gARG is?
> 
> That's the distinction between a GARG and a sample resolved GARG. You can express a GARG without defining any samples. Then all you have is the local parent-child relationships. If you define a set of samples, you can then resolve the ancestry wrt to that set.

But you can have a sample-resolved gARG which defines the edges as (0, L] or (0,x] or (x, L), and you can also have a sample-resolved gARG which defines them so that there are no hanging edges in the local trees. I think both are valid sample resolved gARGs, from your definition above, unless I've missed something.

jeromekelleher:
> But you can have an implicit sample-resolved gARG which defines the edges as (0, L] or (0,x] or (x, L), and you can also have an explicit sample-resolved gARG which defines them so that there are no hanging edges in the local trees. I think both are valid sample resolved gARGs, from your definition above, unless I've missed something.

No, you can only define "hanging edges" in the trees if you define some set of samples. If you don't define a set of samples, then every node in the graph must be consider a sample, and therefore (by definition) you can't have any hanging edges.

There's no implicit-vs-explicit distinction here, the important thing is whether you define some of the nodes as samples or not (and then percolate the ancestry). You could have "samples-but-not-resolved" as well, but what's the point?

jeromekelleher:
> True, but sigma: N -> N_0 only needs to be defined for N (the set of all nodes) rather than N (all the natural numbers)

Like I said, this isn't about practical implementation. It's mathematically more economical to avoid having to define what the set of nodes is.

@JereKoskela - are you comfortable with this or am I on my own here?

hyanwong:
> No, you can only define "hanging edges" in the trees if you define some set of samples. If you don't define a set of samples, then every node in the graph must be consider a sample, and therefore (by definition) you can't have any hanging edges.

I think I misunderstood what you meant by "through the local trees".

> There's no implicit-vs-explicit distinction here, the important thing is whether you define some of the nodes as samples or not (and then percolate the ancestry). You could have "samples-but-not-resolved" as well, but what's the point?

Yes, I thought you meant "samples-but-not-resolved", sorry. 

P.s. did @JereKoskela  say that there was a standard symbol we could return for "no breakpoint", rather than 0?

hyanwong:
By "set of disjoint intervals over which c inherits from p", I thought you meant *locally* (implicit), rather than w.r.t. the samples. Perhaps that needs more clarification?

Does a (non-sample-resolved) gARG require those constraints (that the edges above a recombination node sum up to fill the entire interval)? I think maybe... not?

JereKoskela:
> Like I said, this isn't about practical implementation. It's mathematically more economical to avoid having to define what the set of nodes is.

I don't see a problem with sigma being defined on a domain which is bigger than what we actually need. As long as it's well-defined for each argument we actually want to plug in, it's good in my book.

> a standard symbol we could return for "no breakpoint", rather than 0?

\Delta or \partial are common notations for so-called cemetery states for transient stochastic processes. We could try to co-opt one (probably \Delta) for our purposes as a kind of null state. I'd prefer a symbol which has no independent meaning so that it's clearly arbitrary and we don't intend to do anything with it, but I'm pretty indifferent about what that symbol should be. 

> Then the codomain of sigma becomes more complicated. This is more elegant because a breakpoint at 0 means the interval to the left is empty.

On the other hand, this sounds like a good reason for a symbol with some independent meaning.

hyanwong:
> Then the codomain of sigma becomes more complicated. This is more elegant because a breakpoint at 0 means the interval to the left is empty.

OK, I guess we need to say that X is a positive real then?

hyanwong:
Off the back of https://github.com/tskit-dev/what-is-an-arg-paper/issues/56#issuecomment-1040717922, do we not (normally) define the eARG and the gARG with respect to a set of samples, such that the edges extend backwards from the samples? Perhaps that's not necessary, but I think it is why the structure is called the *ancestral* recombination graph, rather than just the "recombination graph", and it's certainly in the spirit of the original formulations.

The more general structure where *all* lines of ancestry are traced, both forwards and backwards in time, is perhaps the "genetic pedigree", or simply the "recombination graph"?

There's a difference between a "sample-based gARG" and a "sample-*resolved* gARG", though. It's the resolution stuff that implies running the Hudson/simplification algorithm.

jeromekelleher:
This confusion is probably my fault - I'm really talking literally about the encoding. The EARG  and GARG encodings are local and don't refer to a set of samples. So, you could, say, add a new sample at the leaves, or decide that one of the existing internal nodes was a sample, and this wouldn't alter any of the rest of the graph. The "sample resolved GARG" is not local: it's always wrt to some set of samples, and changing the samples will effect other parts of the graph.

So it's not about the model or the process or anything, just literally about the lists of tuples that are used to encode the structure.

argutils/ancestry.py
```diff
@@ -572,3 +573,127 @@ def ca(*args):
...
+            print("parent = ", p)
+            print("Ip = ", I[p])
+            print("Ic = ", I[c])
+            print("e = ", e[2])
+            print("intersection = ", I[c].intersection(e[2]))
+            I[p] = I[p].union(I[c].intersection(e[2]))
+            N.add(p)
+    return [(c, p, I[c]) for c, p, _ in E]
+
+def as_earg(ts):
```
jeromekelleher, line 637: These two functions might help clarify things @hyanwong - this is a literal implementation of the EARG definition and a reasonable method for recovering a local tree from it.

(I'm still working on the GARG and resolved GARG versions)

argutils/ancestry.py
```diff
@@ -572,3 +573,127 @@ def ca(*args):
...
+    Given a minimal EARG definition (E, sigma), return the tree for a the given
+    set of samples at the specified position as a dictionary parent->child.
+    """
+    N = set(S)
+    parent = {}
+    while len(N) > 0:
+        c = N.pop()
+        P = [e[1] for e in E if e[0] == c]
+        if len(P) > 0:
+            if x < sigma[c]:
```
jeromekelleher, line 675: Note: this is why I like defining sigma(u) = inf for non recombination nodes. This is about as cleanly as you can implement this operation.

hyanwong:
> The EARG and GARG encodings are local and don't refer to a set of samples. So, you could, say, add a new sample at the leaves, or decide that one of the existing internal nodes was a sample, and this wouldn't alter any of the rest of the graph. 

Yes, ISWYM. I hadn't noticed the lack of the reverse-time approach in the original definitions. I just thought it worth pointing it out that the encoding that you define will be appropriate for any genome pedigree graph, not just the (G/E)ARG.

> The "sample resolved GARG" is not local: it's always wrt to some set of samples, and changing the samples will affect other parts of the graph.

That's a very good point.

argutils/ancestry.py
```diff
@@ -572,3 +573,127 @@ def ca(*args):
...
+        edge = next(edges, None)
+    # We could wrap sigma as a function to be literal about the definition
+    # but this is simpler for debugging.
+    return E, sigma
+
+
+def earg_get_tree(E, sigma, S, x):
+    """
+    Given a minimal EARG definition (E, sigma), return the tree for a the given
+    set of samples at the specified position as a dictionary parent->child.
```
jeromekelleher, line 667: Note to self - comments are wrong dictionary is mapping child->parent

argutils/ancestry.py
hyanwong: I can see why, computationally, you might want to use `inf`, but that's not the same as using it in the mathematical definition, is it?

JereKoskela: Yan beat me to it, but I was going to make the same point. I'm not strongly opposed to using inf in the text, but it always stands out to me as a bit odd when looking through the draft.

jeromekelleher: Let's see how it goes - I have some draft text describing how we build the trees in the text, and it does help make that description simpler and more concise. We may or may not want to include this. Maybe it's *too* neat though! Anyway, that's my motivation - let's leave it for now.

hyanwong:
A minor point, but given the comments above about the mathematical description specifying the encoding rather than the topology of the ARG , I wonder if instead of saying "we define an eARG as ...", we should say "we encode an eARG as..."? Personally, I think of the *definition* as also including restrictions on the topology, whereas the encoding does not have to.

jeromekelleher:
All of the requirements should be present in the definition - if they're not, it's not a full definition. An encoding is just a way of being mathematically precise about what we're talking about. The encoding absolutely does define the topology - that's the whole point.

hyanwong:
> All of the requirements should be present in the definition - if they're not, it's not a full definition. An encoding is just a way of being mathematically precise about what we're talking about. The encoding absolutely does define the topology - that's the whole point.

A (standard Griffiths) ARG defines sample nodes (or more precisely sampling events), and the topology is defined above those. That's not captured in your definition of an eARG?

hyanwong:
We also need to specify that this is directed (perhaps a given, since we talk about parents and child nodes), and that we can't have cycles, right. We could just say it is a type of DAG.

jeromekelleher:
> A (standard Griffiths) ARG defines sample nodes (or more precisely sampling events), and the topology is defined above those.

Have you got a reference for this? I've struggled to find anything precise.

hyanwong:
> > A (standard Griffiths) ARG defines sample nodes (or more precisely sampling events), and the topology is defined above those.
> 
> Have you got a reference for this? I've struggled to find anything precise.

Griffiths & Majoram define the ARG as the ancestry, backwards in time, of a sample of `n` genes at time t=0. So that's equivalent, right?

JereKoskela:
> We also need to specify that this is directed (perhaps a given, since we talk about parents and child nodes), and that we can't have cycles, right. We could just say it is a type of DAG.

I don't know whether this is usually present in ARG definitions, but the coming down from infinity property of the Kingman coalescent implies that it is possible to construct an ARG for a whole, countably infinite population. I guess you could view that as sampling everyone, but I think that's only a semantic difference from modelling a whole population without sampling as such.

jeromekelleher:
> Griffiths & Majoram define the ARG as the ancestry, backwards in time, of a sample of n genes at time t=0. So that's equivalent, right?

No, I'm talking about the *structure* here. 

Can we leave this discussion until I have some draft text that's ready for review please?

hyanwong:
> > A (standard Griffiths) ARG defines sample nodes (or more precisely sampling events), and the topology is defined above those. That's not captured in your definition of an eARG?
> 
> I don't know whether this is usually present in ARG definitions, but the coming down from infinity property of the Kingman coalescent implies that it is possible to construct an ARG for a whole, countably infinite population. I guess you could view that as sampling everyone, but I think that's only a semantic difference from modelling a whole population without sampling as such.

Yes, by "sample", it could be the entire population. But actually it's not the "sampling" itself that I'm thinking about, it's the fact that the ARG is defined as the ancestry backwards in time from a set of specified nodes. I imagine that you could build a recombination graph that includes nodes that are not ancestral to the specified nodes. That would not, I think, be an ARG?

hyanwong:
> Can we leave this discussion until I have some draft text that's ready for review please?

Yep - sorry for the noise. Let's just wait. Just copying one of my comments above in case I forget too:

> Also, there is a constraint on the eARG that says that there only two parents per child

jeromekelleher:
I've done a bunch of stuff here, and I'm no closer to fixing the notation etc, so it's probably best to just merge this much. Definitely making forward progress though.

hyanwong:
This looks good, and I think we should merge, then change.

My thinking on event ARGs vs general event-based recombination graphs has clarified a little here, and perhaps it would be better continuing the discussion in a specific issue after merging?

jeromekelleher:
> My thinking on event ARGs vs general event-based recombination graphs has clarified a little here, and perhaps it would be better continuing the discussion in a specific issue after merging?

Sure

PR #110, opened 2022-02-14
Use W-F simulator for fig 5

hyanwong:
See https://github.com/tskit-dev/what-is-an-arg-paper/issues/82#issuecomment-1039268250

There's an interesting case here where in (c) we haven't removed the ARG nodes in which multiple lines come together (i.e. node `I`).

~Annoyingly, I just realised, however, that this *doesn't* have any nodes that are both recombination and ca/coalescence nodes (apart from `J` which gets simplified away), so I might need to hunt for another example.~ Actually, node `E` is one of the 2 nodes that contribute to a "recombinant" and also a coalescent node. But maybe we want an example where a node is both a coalescent node and a *recombinant*.

Screenshot 2022-02-14 at 19 04 29

### DIFF
--- a/illustrations/ARG_recomb_node_deletion.py
+++ b/illustrations/ARG_recomb_node_deletion.py
@@ -2,112 +2,91 @@ import collections
 import io
 import itertools
 import sys
+import string
 from pathlib import Path
 
-
-import msprime
-import networkx as nx
+import matplotlib as mpl
+import matplotlib.pyplot as plt
 import numpy as np
 import tskit
-import matplotlib.pyplot as plt
+
+current_dir = Path(__file__).parent
+
+# useful modules in the top level dir: we should be able to import this if the script is
+# run from the top level dir, but for some reason it fails, so we just add the path here
+sys.path.append(str(current_dir.parent.absolute()))
+import argutils
 
 outfile = "ARG_recomb_node_deletion"
 
-# useful scripts in the top level dir
-sys.path.append(str(Path(__file__).parent.parent.absolute() / "utils"))
-import convert
-import ts_process
-import argdraw
+def arg_node_simplification():
+    seed = 372
+    t_x={9: 3, 6: -8, 18: 50, 19: -15, 17: 5, 13: 10, 14: 2, 10: -10, 15: 5, 16: -5, 7: -5, 11: 5, 8: 8}
 
-
-def positions_from_ts(ts, pos=None, return_graph=False):
-    ts_arg = ts_process.flag_unary_nodes(ts)
-    G = ts_process.to_networkx_graph(ts_arg)
-    if pos is None:
-        pos = argdraw.nx_get_dot_pos(G, add_invisibles=False)
-    return (pos, G) if return_graph else pos
+    ts = argutils.sim_wright_fisher(2, 10, 100, recomb_proba=0.1, seed=seed)
+    labels = {i: string.ascii_uppercase[i] for i in range(len(string.ascii_uppercase))}
+    # relabel the nodes to get samples reading A B C D
+    labels.update({2: 'B', 3: 'C', 1: "D"})
+    ts = argutils.viz.label_nodes(ts, labels)
     
-
-def plot_ts_arg(ts, ax, labels=None, pos=None, arrows=False):
-    pos, G = positions_from_ts(ts, pos, return_graph=True)
-    col_map = argdraw.nx_ts_colour_map(G)
-
-    argdraw.nx_draw_with_curved_multi_edges(
-        G, pos, col_map, curve_scale=30, ax=ax, labels=labels)
-
-# SIMULATE
-ts = ts_process.convert_to_single_rec_node(
-    msprime.sim_ancestry(
-        3,
-        sequence_length=1e4,
-        population_size=1e4,
-        recombination_rate=1e-8,
-        record_full_arg=True, 
-        random_seed=76, # by trial and error, a seed of 76 gives a nice small example
+    fig, (ax1, ax2, ax3, ax4) = plt.subplots(1, 4, figsize=(12,5))
+    ax1.set_title("(a) little ARG")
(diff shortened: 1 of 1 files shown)


hyanwong:
Here's another one, swapping

```
seed = 372
t_x={18: 30, 9: 7, 19: -10, 12: 5, 17: 5, 13: 8, 14: 4, 10: -10}
```

Not sure what I think about this. The fully simplified version is quite messy

ARG_recomb_node_deletion

jeromekelleher:
I really like the last e.g.. Interesting the way nearly all the recent stuff survives simplification but the older topology disappears.

hyanwong:
> I really like the last e.g.. Interesting the way nearly all the recent stuff survives simplification but the older topology disappears.

That's because D always remains the outgroup, but just attaches into different places. It gives a good insight as to why there is often a lot of node jumping at the top of the tree, I think (long branches there that provide a lot of scope for SPRs.

hyanwong:
Also interesting that there is a polytomy in the trees, because it is a WF model.

jeromekelleher:
Shall we merge?

hyanwong:
I'll change it to the updated version (seed 372), shall I, then merge? We aren't to worried that the fully simplified version is quite busy?

jeromekelleher:
I think the business is fine - this is a real property of fully simplified graphs. Merge away whenever.

hyanwong:
I'll wait for https://github.com/tskit-dev/what-is-an-arg-paper/pull/85 to be fixed and merged, as I assume we want to show the line weights here too.

Do we want to show line weights on the trees, do you think, @jeromekelleher ?

hyanwong:
What we aren't showing here is the "hanging topology", because we don't have a `resolve=False` argument for that function (see https://github.com/tskit-dev/what-is-an-arg-paper/pull/85#issuecomment-1036089104). Do we want to show this in this figure, or are we past that stage by this point in the paper?

P.s. the diamond is reflected in the last 2 trees in the sequence, which pass though `O` vs `N` respectively, but are otherwise identical.

jeromekelleher:
> I'll wait for #85 to be fixed and merged, as I assume we want to show the line weights here too.

We could just merge and update, but I don't mind.

> Do we want to show line weights on the trees, do you think, @jeromekelleher ?

No, I think this would be confusing.

> What we aren't showing here is the "hanging topology", because we don't have a resolve=False argument for that function (see #85 (comment)). Do we want to show this in this figure, or are we past that stage by this point in the paper?

That's a good point. If we did show this, then I think we'd need to also show the resolved trees (corresponding to c or d, I guess), so would need another row. Also the trees would be very wide I think, might not be practical.

It's probably easy enough the hack an "unresolved" version from the output, if we really wanted it.

hyanwong:
> > I'll wait for #85 to be fixed and merged, as I assume we want to show the line weights here too.
> 
> We could just merge and update, but I don't mind.

Yes, although the best positioning might change if the lines are of different widths (not if we use alpha to show them, though), so I was going to wait and tweak if necessary before merging.

> > Do we want to show line weights on the trees, do you think, @jeromekelleher ?
> 
> No, I think this would be confusing.

I agree.

> > What we aren't showing here is the "hanging topology"...

> That's a good point. If we did show this, then I think we'd need to also show the resolved trees (corresponding to c or d, I guess), so would need another row. 

Do we show that anywhere? You said it clarified things for you, but it might be less obvious to others? It belongs more in the "intrinsic vs extrinsic" figure, if anywhere.

hyanwong:
> It belongs more in the "intrinsic vs extrinsic" figure, if anywhere.

Commented at https://github.com/tskit-dev/what-is-an-arg-paper/issues/56#issuecomment-1040216950 - I don't think we need hanging topology here.

hyanwong:
Interesting with line thickness:

image

jeromekelleher:
Looks great, I'd say merge ASAP.

Do we need to normalise the line-thicknesses or something? Lines that should be the same (say, F-D) look thinner in (d) than (c) to me.

hyanwong:
> Do we need to normalise the line-thicknesses or something? Lines that should be the same (say, F-D) look thinner in (d) than (c) to me.

Good spot. It's a bit weird. It is accurately reflecting the tree sequence edges: here they are printed directly from the tree sequences.

```
graph a, (F - C), Edge(left=0.0, right=40.0, parent=5, child=2, metadata=b'', id=6)
graph a, (F - C), Edge(left=0.0, right=40.0, parent=5, child=2, metadata=b'', id=6)
graph b, (F - C), Edge(left=0.0, right=40.0, parent=5, child=2, metadata=b'', id=3)
graph c, (F - C), Edge(left=0.0, right=40.0, parent=5, child=2, metadata=b'', id=3)
graph d, (F - C), Edge(left=0.0, right=40.0, parent=5, child=2, metadata=b'', id=3)

graph a, (F - D), Edge(left=0.0, right=40.0, parent=5, child=3, metadata=b'', id=7)
graph a, (F - D), Edge(left=40.0, right=61.0, parent=5, child=3, metadata=b'', id=8)
graph b, (F - D), Edge(left=0.0, right=61.0, parent=5, child=3, metadata=b'', id=4)
graph c, (F - D), Edge(left=0.0, right=61.0, parent=5, child=3, metadata=b'', id=4)
graph d, (F - D), Edge(left=0.0, right=40.0, parent=5, child=3, metadata=b'', id=4)
```

You can see that F-C is the same width in all graphs, but F-D spans 40 units in (d), rather than 61 in all the others. I wasn't expecting that...

hyanwong:
> You can see that F-C is the same width in all graphs, but F-D spans 40 units in (d), rather than 61 in all the others. I wasn't expecting that...

Oh, hang on, there's some of the ancestry that goes directly to node G from D, not via F. So that's where it's gone.

hyanwong:
OK, I think this is OK now. Shall I merge?

PR #120, opened 2022-02-17
Tweaks to Figure 4 (added genome pictures to nodes)

a-ignatieva:
What do people think?
ARG_edge_annotations


hyanwong:
Oooh, that's cool. Well done for getting that to work. Green is fully coalesced, right? That means any tree is a local root in the simplified TS.

I like the general idea. I don't know if it's too small to see at this scale? What is the appropriate interpretation in plot (a)?

P.s. they look like hamburgers :)

hyanwong:
> What is the appropriate interpretation in plot (a)?

Should we, for instance, have them all black in (a), apart from the "recombination edges"? With them being clear like this, it invites comparison with the "empty" node in plot (b).

Do we need to refer to the node letters anywhere in the description, or can we do without them?

jeromekelleher:
This is great @a-ignatieva, I like it!

Here's some thoughts:

- It would be good to keep the node labels as we want to discuss particular cases. (We might even want to display the marginal trees). That gives a bit of a placement problem, but it's probably best to locate the labels case-by-case
- I like the hamburgers on the right! :drool:
- I'm less convinced showing the genomes is helpful on the left. The breakpoints are super clear from (b). How would it look if we just left the patties out in (a)?

jeromekelleher:
> It is nice to show the amount transmitted in (a) but that is done by the edge annotations, and I'm not sure what it means to fill out amounts of material on the nodes in (a) anyway.

Exactly - I fear this would confuse people.

a-ignatieva:
Agree about everything! Here's an update: 
ARG_edge_annotations

hyanwong:
This looks great. I think we should merge.

The only thing I'm not sure about is the green colour scheme, but I don't think I have a better solution. By the way, should J have a grey pattie?

hyanwong:
> I don't think I have a better solution. 

Actually, I wonder what it would look like if the samples all started out at 33% grey, then when they merge once they get to 66% grey, then when they merge again (to make the root) they go all black? That might capture the gradual accumulation of samples in each tree.

hyanwong:
> Can mock that up - what's the colour for non-ancestral material then? White?

Yes, I guess so.

Another possibility is to invert stuff and have grey burger buns instead, but that could get confusing. I'll let you see what works.

a-ignatieva:
ARG_edge_annotations

hyanwong:
I quite like the way it builds up to the filled up burger. Is it misleading, do you think?

jeromekelleher:
I really like this, I think it's great.  We could perhaps tweak the grey-levels a bit to make the junctions in the patties clearer, but that's very much nitpicking.

Shall we merge this @a-ignatieva?

a-ignatieva:
Ok great - yep good to go then!

jeromekelleher:
Done - looks great in the document!

PR #122, opened 2022-02-18
Add separate simplification section

jeromekelleher:
Simplification is too big a thing to mush together with ancestry resolution. Making a separate section for it.

WIP


jeromekelleher:
Some text on simplification - might as well merge this much.

PR #134, opened 2022-02-24
Change fig 2 to compare node-vs-edge annotations

jeromekelleher:
I've been thinking a lot about Fig 2, and how we're building the narrative from the familiar Griffiths ARG to the GARG that we want to work with. In the interest of making the smallest changes that we can at once, maybe it's better to talk about the (crucial) distinction between node and edge annotations there (and the corresponding section) and to then move on to talk about GARGs with two parents per recomb in the next section (where it's natural in the context of the pedigree).

This seems like a smoother path to me, and less likely to trip people up by changing too much at once. Here's a (very very!) rough update to what the figure might look like:

node-vs-edge-annotations

Part of the motivation for this is that actually there's no reason that we *have* to use two parents per recombination, and if that's what the output of things like ARGweaver contains, then maybe we should just convert that faithfully. Yes, we will lose information about the breakpoints in diamonds, but I don't think that's much of a loss in practise.

So, I think maybe we just have to bring the reader through the full journey here and perhaps trying to avoid the issue of discussing node vs edge annotations would be a mistake.

Any thoughts?

(Pinging @a-ignatieva and @jerekoskela as I see you're not watching this repo, so I guess you don't get notified of PRs and issues unless I do?)

ps. This is a very rough outline of the figure - if we settle on something like this then we can figure out a better way of displaying the information.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -587,7 +587,7 @@ An example is given in Fig.~\ref{fig-arg-data-structure}A.
 
 \begin{figure}
 \centering
-\begin{tabular}{cc}
+% \begin{tabular}{cc}
 \begin{tikzpicture}[x=5mm, y=5mm, node distance=2mm and 20mm]
 \tikzset{greynode/.style={circle,fill,inner sep=1},
 nodelabel/.style={font=\footnotesize}}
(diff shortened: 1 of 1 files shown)


hyanwong:
I think this is a very good observation Jerome. Your new figure above makes lots of sense, and (one of) the rationales behind the 2-recombination-node version is indeed much easier to see from the pedigree diagram. I say go for it.

hyanwong:
Do we also want to hint in the caption that the edge version allows a node with >2 parents. I think probably we don't, but if we think that's a useful idea, because it tees up the concepts involved in simplification, we could put a forward reference into the caption somehow.

JereKoskela:
I think the figure looks great, and the approach sounds sensible. The leftmost 2/3 of the figure still looks like a familiar ARG, which should make the paper very pedagogical.

For the same reason, I'd avoid bringing up the >2 parents point. We should definitely refer back to Fig 2 when it comes up later to explain that edge annotation is what makes it possible, but a forward reference would be a big leap for the uninitiated reader.

jeromekelleher:
> Do we also want to hint in the caption that the edge version allows a node with >2 parents?

That might be pushing people a bit to quickly, let's take it nice and slow.

jeromekelleher:
Let's merge this so.

hyanwong:
I can work on the changes to introduce 2 parents per recombination in the pedigree figure.

jeromekelleher:
> I can work on the changes to introduce 2 parents per recombination in the pedigree figure.

Sounds good.

PR #135, opened 2022-02-28
Light edit on start of EARG section

jeromekelleher:
Minor text update

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -531,59 +531,40 @@ of inferring parameters of the
 stochastic process, where the ancestry is regarded as a
 latent parameter to be averaged out
 \citep[e.g.][]{griffiths1996ancestral,kuhner2000maximum, nielsen2000estimation,
-fearnhead2001estimating}. These methods met with limited success
+fearnhead2001estimating}.
+% FIXME: need some proper statistical language here to describe why these
+% methods didn't work so well.
+These methods met with limited success
 due to the overwhelmingly large and
-horribly structured [Jere: what's the right way to say this?]
-state space of ARGs. The focus subsequently shifted to
-the more tractable---but still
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -531,60 +531,41 @@ \section*{The Event ARG data structure}
...
+since parsimonious ARGs have no immediate connection to
+the stochastic process. Indeed, the ARG instances produced by
+the coalescent with recombination are usually highly unparsimonious,
+since they contain many invisible recombinations
+(although those produced under the SMC may be parsimonious, depending
+on the pattern of mutations that is present).
 
 % What do we mean by an EARG, informally?
 The Griffiths ARG encoding is an elegant and mathematically
-economical description of the coalescent with recombination on a single
```
jeromekelleher, line 550: @hyanwong - I removed "on a single chromosome" here as I think it would just confuse people. The stochastic process is inherently defined on a single chr.

jeromekelleher:
Minor edits here, will merge in case others are working on the text.

paper.tex
JereKoskela: This isn't completely true. There are natural ways to define an ARG on multiple chromosomes. In Kingman-like cases that amounts to running a single-chromosome ARG independently on each chromosome, but things like [multiple mergers](https://www.sciencedirect.com/science/article/pii/S0025556418303523) or [selection](https://www.sciencedirect.com/science/article/pii/S0040580902000497) break that simplification.

I think we should only cover the single-chromosome case here to avoid packaging in any more nuances than we already have. But in the interest of being as model-agnostic as possible, we should have a brief statement somewhere explaining that that's what we're doing.

jeromekelleher: SGTM - I'll leave it with you @JereKoskela!

PR #140, opened 2022-03-02
Mention optional 3rd change to make a 2-recombination node ARG

### LINKED ISSUE #128: Main 2 changes to the Griffiths ARG, and explaining the 2 node convention
In the text we say

> Fortunately, it is straightforward to generalise the classical Griffiths encoding to both simplify implementations and to encompass any possible pattern of genetic inheritance. We require only two changes. Firstly, we let nodes in the graph model genomes rather than events. As we see in the next section, this allows us to model complex patterns of inheritance without classifying different types of node. The second change we must make is to associate the details of inheritance intervals with edges rather than as breakpoints associated with recombination events and edge ordering rules.

I agree completely with this. We should possibly add a 3rd, (IMO more minor) change, which is that by convention, we also strongly suggest that recombinations are encoded using 2 genomes (nodes) rather than one (Fig 2). I would like to point out that this isn't strictly necessary, for a gARG (hence the first 2 changes are the most important), but it makes simulating the ARG much easier, and more importantly, it means that we can merge edges without changing the meaning of the graph (i.e. we can store diamonds without requiring multiple edges).

However, the caption for Fig 2 seems not quite to follow this train of thought. It says

> ... The most important change between the two encodings is to represent a recombination event (node 3 in A) by two parent genomes (nodes 3 and 4 in B) and explicitly storing the interval that the child inherits from each parent along the two corresponding edges. 

I would probably put the interval storing first, and rephrase a little.

hyanwong:
Fixes #128 with the rough text Jerome suggested.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -810,7 +810,11 @@ patterns of inheritance without classifying different types of node.
 The second change we must make is to associate
 the details of inheritance intervals with \emph{edges} rather than
 as breakpoints associated with recombination events and
-edge ordering rules.
+edge ordering rules. A third (technically optional, but practically useful)
+change is to describe recombinations in terms of two parental genomes
+rather than a single node representing either the recombination event or
+the gamete in which recombination has occurred.
+
 % That is, given two genomes $p$
 % and $c$ where $p$ is the parent genome and $c$ is the child
 % (although many generations may have elapsed between the two),


jeromekelleher:
I don't think we want to say this yet - the changes in #138 are part of the switch to making this section *just* about the edge-vs-node annotations (which is the fundamentally important thing, after all). The next section then will talk about how it's simpler if we think about genomes rather than events, particularly in the light of modern datasets with, e.g., deeply sequenced complete pedigrees, and individuals with hundreds of thousands of offspring (e.g. a good AI bull).

hyanwong:
Yes, I wasn't sure, but thought it was worth having it as a PR so that we didn't forget the issue. It depends on the flow of the argument.

jeromekelleher:
Shall we close this so?

hyanwong:
I've got some ideas for changes that will close this. Leave it open until then - should be tonight.

PR #141, opened 2022-03-02
Changes for fig3 caption

hyanwong:
I think this describes what is now shown in fig 3.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -861,16 +861,17 @@ in a pedigree. (a) Diploid individuals, visualised in a highly inbred pedigree a
 labelled $i_1$ to $i_8$, contain both paternal (left) and maternal (right) genomes
 labelled \textsf{A} to \textsf{P}. Black lines show transmission paths connecting
 genomes in the current generation (\textsf{A} to \textsf{D}) with their ancestors.
-Two recombination breakpoints, at positions 3 and 6 mean that genomes \textsf{A}
+Two recombination breakpoints, at positions 2 and 7 mean that genomes \textsf{A}
 and \textsf{C} are independent recombinants of the paternal genomes \textsf{E}
 and \textsf{F}: the region of genome transmitted through each edge is shown in
 these cases. Where genomes merge into a common ancestor, the merged region
 is shown darker, until all genomic regions are fully coalesced in genome \textsf{N}
+in the oldest generation.
 (b) The equivalent genome ARG, where there is a single type of node (representing
 a genome) along with annotations on all edges; annotations showing partial
 transmission are highlighted in bold. Recombination, such as that causing the
-breakpoint at position 3, is associated with two parent genomes (\textsf{E} and 
-\textsf{F}) and a child genome (the recombinant, \textsf{A}).
+breakpoint at position 7, is associated with two parent genomes (\textsf{E} and 
+\textsf{F}) and a child genome (the recombinant, \textsf{C}).
 (c) The edge-annotated genome ARG in (b) fully defines the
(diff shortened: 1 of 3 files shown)


paper.tex
```diff
@@ -856,14 +856,11 @@ \section*{Genome ARGs}
...
... visualised in a highly inbred pedigree and labelled $i_1$ to $i_8$, contain both paternal (left) and maternal (right) genomes labelled \textsf{A} to \textsf{P}. Black lines show transmission paths connecting genomes in the current generation (\textsf{A} to \textsf{D}) with their ancestors. Two recombination breakpoints, at positions 3 and 6 mean that genomes \textsf{A} and \textsf{C} are independent recombinants of the paternal genomes \textsf{E} and \textsf{F}: the region of genome transmitted is shown in these cases. Where genomes merge into a common ancestor, the merged region is shown darker, until all genomic regions are fully coalesced in genome \textsf{N} (b) The equivalent genome ARG, where there is a single type of node (representing a genome) along with annotations on all edges. Recombination, such that causing the breakpoint at position 3, is associated with two parent genomes (\textsf{E} and  \textsf{F}) and a child genome (the recombinant,  \textsf{A}). In this case two independent recombinations share the same parent genomes, because the individuals containing the recombinants are siblings. Edge-annotation allows this complex genetic relationship to be fully encoded.
```
jeromekelleher: Can you line break this at ~80 spaces please?

jeromekelleher: > Recombination, such that causing the breakpoint at 

this doesn't quite parse?

jeromekelleher: I don't think we need the two sentences starting "In this case two .."

paper.tex
```diff
@@ -856,14 +856,11 @@ \section*{Genome ARGs}
...
+% TODO - do we want to mention that by allowing multiple edges between nodes, we can
```
jeromekelleher, line 861: No - I think we'll have covered that in the previous section. PS we don't want to say "multiple edges between nodes" - it's just associating a *set* of intervals with an edge rather than one interval. Let's not get into the graph-vs-multigraph loop again.

jeromekelleher (commented):
LGTM, minor comments

paper.tex
hyanwong: Oh yes, sorry - my latex editor wraps it so I didn't spot.

hyanwong: Typo, should be "such as that"

hyanwong: > I don't think we need the two sentences starting "In this case two .."

I was trying to show what a complex relationship looks like.

Happy to delete, but this is the point where we introduce the idea of 2 "recombination nodes" + 1 "recombinant" per recombination, right? The hidden idea here is that in a WF model we can't simply store recombination points on the parents, because parents can be shared. But maybe we just say that in the main text?

paper.tex
hyanwong: Yeah, I wasn't sure, hence the TODO. I think we do want to imply that GC etc can be captured here, but as you say, the main text should cover that.

Re graph/multigraph, I was only using the tskit terminology as a shorthand. "Multiple tskit edges between nodes" would have been more accurate, you are right.

paper.tex
jeromekelleher: I think people will get that. Don't want the caption to be too long.

Can you do the line breaks please? Makes it very difficult to comment on specific bits with one long line.

hyanwong:
Thanks for the feedback, merging

paper.tex
hyanwong: Done, thanks.

PR #146, opened 2022-03-03
Clarified single chr scope and unified Fig1 with text

### LINKED ISSUE #136: Brief comment on single vs multiple chromosomes
As discussed in #135, it would be good to find a place for a brief explanation of the fact that it is sometimes necessary to define explicitly multi-chromosome ARGs, but that we're only talking about the single-chromosome case for (relative) simplicity. The more obviously reasons for multi-chromosome models include fitness functions with interaction between chromosomes, or multiple merger models which can feature simultaneous mergers in all chromosomes.

I'll put a draft together.

JereKoskela:
A few small updates:

- Added a comment to the end of the intro section explaining that we only focus on discrete genomes and a single chromosome. It's currently a bullet point aimed at going into the end of a roundup paragraph. I think that would be a good place for it: it will tell specialists what to expect, but I think it is still non-technical enough that non-specialists shouldn't be confused. I'm open to opinions on whether it should go elsewhere though, or if it is needed to begin with.
- Slightly tweaked the spacing in Fig 1 to make lines not overlap now that it has been scaled down.
- Made Fig 1 consistent with the notation in the surrounding text (`m` rather than `L` for genome length), and made it work with discrete sites. This makes the recombination rates bulkier, but avoids us having to introduce continuous genomes on the fly in a figure caption. 
- Elaborated on the state space of ARGs, and on for what kinds of genome lengths the CwR is a valid approximation.

Closes #136

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -118,6 +118,11 @@ with some examples.
 % Perhaps we should also mention something about simplification and unknowable
 % nodes, e.g. "show how some of the nodes are not required for useful interpretation
 % of ancestry".
+\item Throughout, we consider ARGs defined for a single chromosome comprising
+of a fixed number of discrete sites. Our argument applies equally well to ARGs
+defined for several non-homologous chromosomes \citep{fearnhead2003ancestral,
+koskela2019robust} and to idealised, continuous chromosomes, but addressing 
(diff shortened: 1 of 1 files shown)


jeromekelleher:
Very nice, thanks @JereKoskela!

PR #150, opened 2022-03-03
Reword the 2 node recombination section

hyanwong:
I don't think we need to back refer to fig 2 any more here, as fig 3 is where we now introduce the 2-RE-node idea. I have also moved the formal definition of a gARG nearer the top of that section. Comments etc welcome - there are many improvements that could be made to my text, I'm sure.

This closes #140

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -908,6 +908,19 @@ topology along with the genome coordinate annotations on the
 edges then uniquely defines the trees at every position
 along the genome (Fig.~\ref{fig-arg-in-pedigree}c).
 
+Formally, we can define a GARG as a set of edges $E$, where each
+element of the set is a tuple $(c, p, I)$ such that $c, p \in \mathbb{N}$,
+$c$ is the child node, $p$ is the parent node and $I$ is the set of
+disjoint genomic intervals $(\ell, r]$ over which genome $c$ inherits from $p$ (we may
+also phrase this in terms of the ancestral material of a sample---see
+the next section). Deriving the local tree at a point $x$
+follows largely the same pattern as for an EARG but is somewhat
+simpler because coordinate information is directly associated with
+edges. Starting with the sample nodes $S$ we traverse
+upwards through the graph from child to parent. At each node $u$, we find an
+edge $(c, p, I) \in E$ such that $u = c$ and $x \in I$ (if it
+exists) and move to the next node $p$.
+
 The focus on modelling genomes rather than events has some interesting
 and important consequences. Firstly, the graph may contain nodes that
 do not correspond to any particular event in the ancestry of the samples.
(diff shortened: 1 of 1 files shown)


hyanwong:
A thought: this doesn't really address the point that the 2-RE-node approach is a technically optional (but practically useful) feature of the gARG as plotted in Fig 3. This is probably the place to do that.

If the single-re-node approach is used, it would be equivalent to placing the gamete as a separate genome, in the same generation as the parents. This may be too much detail to provide here, however (although it may be worth a comment in supplementary material, if we have any)

hyanwong:
Perhaps there are actually 3 things (rather than 2) to mention here:

1. You can have "pass through" type nodes that are not associated with any event (para already written)
2. Nodes can encapsulate multiple events. For example, if node A in fig 3 were to have 2 children, it would be a node with two parents and 2 children. This is quite different from an eARG. Annoyingly that's not actually shown in fig 3, but I can't see any way to do that without adding another generation. Later on we can refer back to this when we start simplifying away recombination nodes, but there's no need to mention that at this point in the text
3. (Optionally), if we chose to use nodes to represent genomes in successive generations, then a recombination is represented by 3 nodes (2 parents and a child), as illustrated by nodes E,F and (say) C. We could, or could not, make the point that the parent nodes may be pass-through nodes (see point 1), although sadly this is not illustrated in this example.

I don't know if we want to change the example at this stage. Adding another recombination might make the whole thing too complicated.

paper.tex
```diff
@@ -921,62 +934,41 @@ \section*{Genome ARGs}
...
-% are shared between siblings, and can also be coalescent nodes
-% this ties in with stuff discussed 3 paras below.
-[More stuff here: clarify why we assign coordinates to edges
+Secondly, the natural way to represent the process of recombination in a
+gARG is not as a single ``recombination node'' (as in an eARG), but with
+two parent genomes between which recombination takes place (e.g.
+nodes \textsf{E} and \textsf{F} in Fig~\ref{fig-arg-in-pedigree}), and a
+single resulting ``recombinant'' genome (e.g. \textsf{C}). The
+edges that link this recombinant with its two parents carry the information
+concerning breakpoint location that would normally be present on the
```
jeromekelleher, line 943: tex quotes

paper.tex
```diff
@@ -921,62 +934,41 @@ \section*{Genome ARGs}
...
+"recombination node" in an eARG. For example, node \textsf{C} in
+Fig~\ref{fig-arg-in-pedigree} inherits genome positions 0 to 7 from \textsf{E}
+and positions 7 to 10 from \textsf{F}, indicating that the breakpoint is at position 7.
+Unlike the eARG, the gARG need not make any distinction
+between common ancestor and recombination nodes: indeed, a
```
jeromekelleher, line 948: two "indeeds" in quick succession here

jeromekelleher (commented):
LGTM, some tiny nitpicks

paper.tex
hyanwong: Good catch, thanks.

jeromekelleher:
Merge away when you're happy

jeromekelleher:
Can you make these changes please @hyanwong so we can merge?

hyanwong:
Yep, sorry, I was rewriting to incorporate the 3 point idea [above](https://github.com/tskit-dev/what-is-an-arg-paper/pull/150#issuecomment-1059037251)

hyanwong:
Merging so we can carry on working on this

PR #151, opened 2022-03-08
Discussion of SPRs

jeromekelleher:
Here's a start on a section discussing the relationship between ARGs and SPRs. I think we have to tackle this, as it's a fundamental property. I also want to address the (casually made, usually) assertion that an ARG is equivalent to a set of trees. What properties to these trees need to have in order to recover the graph?

I think there's two cases:
1. All nodes in the trees are **labelled** (and the labels correspond to the ARG node, obvs): easy
2. Nodes are *not* labelled: requires figuring out what the SPR between each pair of trees *must have been* and reasoning about the node identities based on that. This seems quite hard to me.

Also, unless there's unary nodes stored or some sort of out-of-band info attached, we can never know the exact times of the recombination events.

Does this sound about right?

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1252,6 +1252,8 @@ a given list of
 % want to say "standard phylogenetic" trees here or something, but
 % and they don't have to be binary (just arity > 1), but thought
 % this would be familiar to readers?
+% Jere: Could just say leaf-labelled "marginal trees", "ancestral trees",
+% "local trees", or something like that.
 leaf-labelled binary trees.
 
 [ Cite \citep{deng2021distribution} somewhere here.]
--- a/paper.tex
+++ b/paper.tex
@@ -1185,10 +1185,67 @@ The final level of simplification is to remove \emph{all}
 unary nodes from the local trees (Fig.~\ref{fig-simplification}D
 and E). [And continue]
 
+\section*{ARGs and SPRs}
+\citet{wiuf1999ancestry,wiuf1999recombination} recast
+the coalescent with recombination
+as a process along the genome rather than backwards in time.
+In this formulation we begin at one end of a chromosome by
+sampling a tree under the standard coalescent, and then reason
+about how this tree is modified by recombination events
+as we move along the genome. Although this algorithm is
+significantly more complicated to simulate than Hudson's
+backwards in time approach, it has led to substantial
+insights into the process of recombination itself
+and to the sequentially Markov Coalescent (SMC)
+approximations~\citep{mcvean2005approximating,marjoram2006fast}
+which are central to many modern methods.
 
-[We should discuss the relationship between ARG nodes and SPRs
-here. Cite \citep{deng2021distribution} too. Maybe we want to make
-this a separate section, as it's a subtle and important point.]
+Under the coalescent with recombination (or SMC) each local
+tree along the genome is transformed into the next by
+a ``subtree prune-and-regraft'' (SPR)
+% FIXME there should be earlier refs for SPRs, right?
+operation~\citep{song2003on,song2006properties}.
+The number of SPRs required to transform one leaf-labelled tree
+into another is known as the SPR distance [cite]
+and is computable in O(??) time.
+
+[Discussion of SPRs in an EARG]
+
+The effects of recombination as we move from left-to-right along
+the genome in an unsimplified GARG are easy to interpret.
+Consider the leftmost tree in Fig~\ref{fig-simplification}.
+In the first tree the parent of B is F, and in the second
+tree its parent is E. Since both parents are present as
+unary nodes in the tree, we simply change one edge in
+order to effect the transition.
+A more complex transition occurs when we have a regraft operation
+that occurs above the current root (i.e., the current oldest
+non-unary node). For example, the transition between the
+second and third trees corresponds to node D switching
+from the path through I to H, and we then must add branches
+corresponding to passing through L and meeting at M.
+
+The equivalence between an ARG and a set of local trees
+separated by SPRs is often mentioned [e.g. x y z], but
+it is not obvious how to produce an ARG given (say) a
+collection of newick trees. If the internal nodes are
+labelled, then we can simply input the edges for each
+tree in turn, which can easily be ``squashed'' down
+to a minimal representation. However, internal nodes
+are rarely labelled in newick trees, and therefore
+the process of contructing a graph is much harder.
+In princple, we can contruct an EARG by reasoning
+about the SPRs that must has occured between two
+adjacent trees and postulating the recombination
+nodes that must have occured. However, this process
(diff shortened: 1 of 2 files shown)


JereKoskela:
I think I agree with everything. Just to ruminate on the case of unlabelled nodes for a moment to sanity-check myself: I think the Wiuf & Hein perspective lets us detect from where an SPR between two sites must have detached relatively easily. Is the hard part to work out where it reattaches, because that can (in principle) be anywhere on any other marginal tree, which is a huge and complicated search space?

jeromekelleher:
I'm not particularly sure on this point @JereKoskela, and haven't thought it through properly. It just seemed like it must be a hard problem since computing the SPR distance is NP hard in general (via [this paper](https://link.springer.com/article/10.1007/s00026-001-8006-8) apparently, haven't read it), so even this easier problem must be quite hard. Given that we know the SPR distance is 1 (or at least have assumed it), how hard is it to figure out what that SPR is?

hyanwong:
ISTR that if we *know* there is one SPR, and we have the edge diffs, it is easy to work out what the potential possible SPRs are, but there isn't necessarily a single unique answer. I have done this before in the past.

jeromekelleher:
> and we have the edge diffs, it is easy to work out what the potential possible SPRs are, but there isn't necessarily a single unique answer.

We don't have the edge diffs, that's the point. Our input is two newick trees that we are assured differ by one SPR. How do we charaterise this SPR? We have to solve this problem if we are able to go from an unlabelled list of trees to an ARG.

hyanwong:
> We don't have the edge diffs, that's the point. Our input is two newick trees that we are assured differ by one SPR. How do we charaterise this SPR? We have to solve this problem if we are able to go from an list of trees with unlabelled internal nodes to an ARG.

A large number of SPR changes result in a change in node height but not topology. How would we spot these as different if the nodes are unlabelled? Would we use e.g. node time as a proxy for a label?

jeromekelleher:
> A large number of SPR changes result in a change in node height but not topology. How would we spot these as different if the nodes are unlabelled? Would we use e.g. node time as a proxy for a label?

Good point. That's what we do in practise reading in ms trees and converting to tskit format. There are technical problems, but they are due to newick's use of branch lengths and nothing particularly fundamental. It does assume that you have unique times on internal nodes, though, which is true in the coalescent but not in general.

We've been focusing on topology, and I think we should stick to that. Parsimony ARGs for example won't have times at all. The labelling point is fundamental and it'd be nice to have an answer for that. So, assuming you have unlabelled newick trees with no branch lengths, how hard is it to determine the SPR?

JereKoskela:
I don't think determining the SPR is all that hard. The maximum agreement forest (MAF) between the two trees of interest will consist of 2 trees. Finding the MAF is NP-hard in general, but [Hein et al. (1996)](https://www.sciencedirect.com/science/article/pii/S0166218X96000625?via%3Dihub) have a heuristic which is off by at most a factor of 3. Then the problem reduces to piecing together at most 6 trees into 2, but I would have thought that can't be too hard.

I would expect that the hard part is threading together the trees into some sort of reasonable ARG. If we don't insist on some kind of "reasonableness", then any JBOT is a valid ARG if you just assume that all sites recombine into separate lineages in an epsilon of time.

jeromekelleher:
> I would expect that the hard part is threading together the trees into some sort of reasonable ARG. 

Under the assumptions though there should only be *one* ARG -  the statement that's usually made is "an ARG is equivalent to a list of local trees" and IMO it's not at all as simple as that. Yes, there is a well defined list of local trees from an ARG, but going back from that list of local trees to an ARG seems rather most difficult to me.

So, suppose we have a two-tree ARG, and we've identified the SPR that's occured. How do we construct the Griffiths EARG for this? Well, everything that's not immediately affected by the SPRs will be the same in the two trees, so we can easily build the graph of that. Then, we need to insert the recombination node somewhere. I guess this is the sort of thing that Espalier must be doing, so maybe I should look at that for info.

It does make by brain hurt quite a lot trying to figure out how to do this without internal node labels.

hyanwong:
> the statement that's usually made is "an ARG is equivalent to a list of local trees" and IMO it's not at all as simple as that.

I know I said this, but trivially it can't be true because of ARG changes that don't change the local topologies. There are lots of these that are important to make the (Griffiths) ARG.

JereKoskela:
> Under the assumptions though there should only be _one_ ARG - the statement that's usually made is "an ARG is equivalent to a list of local trees" and IMO it's not at all as simple as that.

I think the only way to make this true is to interpret "equivalent" to mean that all ARGs which are compatible with the list of local trees give rise to the same distribution of genetic diversity in the sample. If you take it to mean that the list of local trees specifies a unique Griffiths EARG, then I think the claim is false. You can add any number of diamonds, or other trivial things like that, without affecting any local trees.

If we're neglecting branch lengths, then you can also always do what I described earlier: recombine every link on each each leaf before any coalescences, whereupon you can freely make any tree sequence you like. Branch lengths would rule that approach out under the CwR because identical node times can only arise in a joint coalescence event, but there will still be similar freedom in parts of the local trees which don't have identical node times.

JereKoskela:
It's probably true that a list of local trees specifies a unique ARG with a minimal number of SPRs, maybe under mild extra constraints to rule out pathologies. But that sounds like it gets us back into NP-hard territory since I think even determining the minimum number of recombinations for SNP data is NP-hard.

jeromekelleher:
Thanks for the input - here's an update. Will need another pass through to get the terminology consistent etc, but does this get the basic point across well enough to merge?

paper.tex
```diff
@@ -1185,10 +1185,76 @@ \section*{ARG simplification}
...
+differ due to the SPR separating adjacent trees in order to
+assign node identity. Without the internal unary nodes marking
+recombinations, we must infer their existence. In
+general, there will not be a unique solution to this problem,
+and so there can be many possible ARGs compatible with
+a given list of
+% want to say "standard phylogenetic" trees here or something, but
+% and they don't have to be binary (just arity > 1), but thought
+% this would be familiar to readers?
+leaf-labelled binary trees.
```
JereKoskela, line 1255: Could just say "marginal trees", "ancestral trees", "local trees", or something like that. Then there is no implication of binarity, but readers should still be on familiar ground.

JereKoskela:
Looks good to me, thanks @jeromekelleher! I've made one comment above, but that's more for future reference rather than something that needs resolving before we merge.

paper.tex
jeromekelleher: I want to contrast with things that contain unary nodes though, which are quite unfamiliar to people I think (indeed, it seems like you can't have them in R's phylo objects)

I've added your note to the text anyway, something to mull for next pass.

Issue #133, opened 2022-02-22
Convert function for kwarg


jeromekelleher:
Following the pattern set for argweaver in #130, we'd like a ``convert_kwarg`` function.

This needs:
- A function ``convert_kwarg(infile)`` in argutils/convert.py. 
- Some example files in the ``examples/`` directory that we can test on.
- A test class TestConvertKwarg in argutils/tests with some very basic tests, just to make sure it's roughly working.

@a-ignatieva - you're the obvious candidate here, fancy picking this up?

jeromekelleher:
Feel free to ping me if you have any qusetions!

a-ignatieva:
I've made a draft PR with what I've got - just need to test that it works properly! Will do that by the end of this week definitely. Sorry for the delay! I've been distracted by world events.

jeromekelleher:
Thanks @a-ignatieva, no worries at all. Hope you're OK.

PR #152, opened 2022-03-08
Convert function for KwARG output

a-ignatieva:
#133 Not thoroughly tested yet! Seems to work on the kreitman example (which I've added to the examples folder). 
kreitman_kwarg

### DIFF
--- a/argutils/convert.py
+++ b/argutils/convert.py
@@ -88,3 +88,46 @@ def convert_argweaver(infile):
     # garg = argutils.earg_to_garg(ts)
 
     return ts.simplify(keep_unary=True)
+
+def convert_kwarg(infile, num_samples, sequence_length):
+    """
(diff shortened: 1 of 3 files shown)


jeromekelleher:
LGTM - I say we merge! @hyanwong, any thoughts?

jeromekelleher:
The graph looks great I think, you can really see the structure that's being inferred. Looks quite similar to what's in one of Yun's papers as well I think, doesn't it?

hyanwong:
Yes, this looks really nice to me. I think we should merge and we can always make improvements after.

hyanwong:
By the way, this is what happens when you run keep_unary_in_coal on the KwARG output. I'm assuming that we don't want to do that, though.

Screenshot 2022-03-10 at 18 09 13

PR #154, opened 2022-03-09
Correct references to figs (a), (b) etc

hyanwong:
For clarity I have switched to lowercase subfigure letters, to avoid confusion with the node labels (which are capital letters). This may eventually need to be changed depending on the journal style.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -357,7 +357,7 @@ description of the algorithm, and how it may be implemented efficiently.
 \end{tikzpicture}
 \end{tabular}
 }
-\caption{(A)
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I was using capitals with a particular journal style in mind (Genetics) where I would hope we submit this. I'd rather not change it.

hyanwong:
All now changed to match Genetics style in https://github.com/tskit-dev/what-is-an-arg-paper/commit/f1b7178861219d8b73d4a59c43414006c41e2ee7

PR #157, opened 2022-03-10
Start on fig 6

hyanwong:
Not by any means perfect, but it's a start.

I'm not sure what labels we want to use for internal nodes. Neatest would be to omit the labels altogether, and simply have small dots (could be coloured depending on node type). But then it would be hard to refer to specific nodes in the text.

This is how it looks for me:
Screenshot 2022-03-10 at 18 57 29


hyanwong:
I am inclined to remove internal node labels because otherwise it invites comparison between the nodes in difference subplots, and the node labels do not at all match. If we want to label specific nodes, we can do it on a case-by-case basis.

hyanwong:
> I am inclined to remove internal node labels

Here's a demo

Screenshot 2022-03-10 at 21 18 35

jeromekelleher:
Looks great! I agree the internal node labels don't add much. Merge whenever your happy.

(BTW the vestigial root on the tsinfer output is a bit embarrassing. Hard to know what the right thing to do is here - if we leave it in, it'll confuse people, if we take it out it could be a bit misleading. What do you think?)

jeromekelleher:
The tsinfer output is definitely good enough to illustrate the main point we want to make, that nodes are *shared* in adjacent trees.

This viz is really useful! Excited to see what Relate and ARGweaver look like now!

jeromekelleher:
Oh yes - it would be good to show how many trees are in each one. I wonder if we could have another row along the bottom showing the position of breakpoints along the genome? That would really help us to see the differences between the methods.

hyanwong:
> Looks great! I agree the internal node labels don't add much. Merge whenever your happy.

Thanks. I'm working on the ARGweaver one now. I'll merge and open issues to improve it.

> (BTW the vestigial root on the tsinfer output is a bit embarrassing. Hard to know what the right thing to do is here - if we leave it in, it'll confuse people, if we take it out it could be a bit misleading. What do you think?)

I think this is the "ultimate ancestor" isn't it? We [say](https://github.com/tskit-dev/tsinfer/blob/6ed592f62d3c92c00eb918ef2bcb50c8c8171bad/tsinfer/inference.py#L952)

```
            # Add the ultimate ancestor. This is an awkward hack really; we don't
            # ever insert this ancestor. The only reason to add it here is that
            # it makes sure that the ancestor IDs we have in the ancestor file are
            # the same as in the ancestor tree sequence. This seems worthwhile.
```

So because we don't use it, perhaps we can delete it?

Actually, now that we save the ancestor ID in the metadata, do we actually need to add the ultimate ancestor ID to the inferred tree sequence at all? What if we simply drop it from tsinfer?

jeromekelleher:
> So because we don't use it, perhaps we can delete it?

Yeah, I think it's probably the simplest thing to remove it from the viz here. This is the "ancestor of the ultimate ancestor" which is just put in for convenience. We probably should just drop it from tsinfer all right.

PR #164, opened 2022-03-11
Add files for ARGweaver inference

### LINKED ISSUE #160: Add ARGweaver output to fig 6
I'm working on this

hyanwong:
Fixes #160. Draft at the moment, as we need to actually run ARGweaver to create an example ARG, then convert it


tools/Makefile
```diff
@@ -0,0 +1,8 @@
+all: | argweaver
+	#the following tools require stand-alone compilation
```
jeromekelleher, line 2: Shouldn't be calling "make" on the local makefile, just 
```
all: argweaver
```
should do it.

tools/Makefile
```diff
@@ -0,0 +1,8 @@
...
+	make -C argweaver
+
+argweaver: 
+	git clone https://github.com/CshlSiepelLab/argweaver
+	# checkout a specific revision (latest as of Mar 2022)
+	cd argweaver && git checkout 7191343c1d43c02de04e4e0a150d77b0612728c6
```
jeromekelleher, line 8: probably needs some compilation as well, or is Argweaver pure python?

hyanwong: It has a Makefile inside it, and when I type `make` in the tools dir, it seems to compile ARGweaver OK. It says

```
...
Note: switching to '7191343c1d43c02de04e4e0a150d77b0612728c6'.
...
#the following tools require stand-alone compilation
make -C argweaver
g++ -c -march=core2 -mtune=haswell ...
```

hyanwong: Oh, hang on, that's what the line above is doing, right?

```
all: argweaver
        #the following tools require stand-alone compilation
        make -C argweaver
```

That's how we did it in `mcveanlab/treeseq-inference`

jeromekelleher: Ah right, that's confusing. Would be clearer to put the ``make -C argweaver`` rule at the end of the argweaver bit.

hyanwong:
I wrote some stuff in the Makefile to create both the files required for input to KwARG, tsinfer & ARGweaver, and to actually run those tools and create the output tree sequences.

I'm sure I haven't done the Makefile right though: first time I've actually tried to write one.

hyanwong:
Given the huge download needed for Relate, I think we should indeed package the Relate, ARGweaver, etc output files (called `Kreitman_SNP_relate.trees`, `Kreitman_SNP_argweaver.trees`, `Kreitman_SNP_tsinfer.trees`) in the repo and not normally run the makefile commands that create them.

jeromekelleher:
Sgtm

jeromekelleher (approved):
LGTM

jeromekelleher:
Don't worry about making the Makefile optimal - we can look at it later if needs be.

hyanwong:
> LGTM

I'll merge then iterate, then

PR #168, opened 2022-03-14
Added sampling probability discussion

JereKoskela:
Added a section on evaluation of sampling probabilities, which amounts to counting lineages and links. I've tried to err on the side of brevity (even though the result is still not very brief), so it would be useful to hear whether this makes sense to people other than me.

### DIFF
--- a/paper.bib
+++ b/paper.bib
@@ -436,6 +436,12 @@
     note={bioRxiv 2021.11.03.466843}
 }
 
+@unpublished{guo2022recombination,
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
Looks good. I'll need to think about the last paragraph more, but how about we merge and iterate?

JereKoskela:
> Looks good. I'll need to think about the last paragraph more, but how about we merge and iterate?

That sounds fine to me.

PR #173, opened 2022-03-18
Add labels to pedigree

hyanwong:
Adds two `(0,10]` labels to the pedigree diagram. Not sure if I like this or not, but having it in means we can say that we show all the edge annotations in the first generation but not upwards from that.

I won't merge yet, but we can leave this as an open PR until we decide which we prefer.


jeromekelleher:
I don't mind, whichever you prefer.

jeromekelleher:
Do you still want to merge this @hyanwong or can we close?

PR #174, opened 2022-03-18
Change fig 2 to letter labels

hyanwong:
Uses `A`, `B`, `C` instead of 1, 2, 3 for fig 2. I'm not sure which I prefer. 1, 2, 3 is perhaps simpler to explain mathematically, but `A` `B` `C` is what we use for the rest of the figures, and labels the nodes using a different scheme from the edges (i.e. distinguishes "0" in the edge column from "0" in the node column).

I'm not sure what the norm is for ARG pictures and for mathematical graphs: possibly numbers?

Not to merge until we decide what we prefer. I had to do this for the ARG talk, so thought I might as well PR it too.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -588,10 +588,10 @@ nodelabel/.style={font=\footnotesize}}
 \node (s6) [greynode] at (3, 4) {};
 
 \node [nodelabel,anchor=north west] at ($(s3) + (0,0)$) {$x$};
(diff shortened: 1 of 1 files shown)


JereKoskela:
I doubt there's strong conventions on the graph side, but I'd guess that letters are more common. They have the advantage of being generic. To me, numbers always feel like the magnitude should mean something.

jeromekelleher:
OK, let's go with this then. Merge away whenever @hyanwong

PR #175, opened 2022-03-18
Add Relate data

hyanwong:
Adds the data, but the conversion routine doesn't work yet


make_example_outputs.py
```diff
@@ -85,9 +90,41 @@ def run_argweaver():
         ts.dump("examples/Kreitman_SNP_argweaver.trees")
 
 
+@click.command()
+def run_kwarg():
```
jeromekelleher, line 94: wrong function name here

jeromekelleher (commented):
I'm still not convinced this is a good idea, and would prefer we used the direct output that Relate produces. We want to illustrate a JBOT, and that's what timed Relate output actually is, so it's not misleading.

I think we should get someone on board who works with this format if we want to use it, rather than try to reverse engineer it.

hyanwong:
> I'm still not convinced this is a good idea, and would prefer we used the direct output that Relate produces. 
> We want to illustrate a JBOT, and that's what timed Relate output actually is, so it's not misleading.

I definitely agree that we should show the Relate tree sequence JBOT, but I think that the fact it is a JBOT is a feature of the conversion to tskit format, not of the relate output itself.

If possible, I think we should *also* show the non-JBOT (undated) form too. Otherwise the (reasonable) accusation would be that we are claiming that Relate is not making a proper ARG, whereas Leo would argue that it is, but that the tree sequence format simply can't represent it properly.

> I think we should get someone on board who works with this format if we want to use it, rather than try to reverse engineer it.

We can email Leo. I do understand the format though.

a-ignatieva:
I agree with Yan - it's not true that Relate itself infers JBOTs (even if that's how they end up looking in the tskit conversion), and it would be good to make that clear.

hyanwong:
Screenshot 2022-03-18 at 17 27 03

jeromekelleher:
OK, sounds good. I guess ideally we'd show the undated and dated Relate output then.

So does Relate do something a bit like [Espalier](https://www.biorxiv.org/content/10.1101/2022.01.17.476639v2) then @a-ignatieva, i.e., reason about internal node identity based on how they partition the samples between adjacent trees?

hyanwong:
Relate reasons about shared edges, but doesn't then extend the logic to shared *nodes*, afaik. It's (relatively) easy to get these shared edges out of the relate format, but I'm wrestling with the logic of then using them to assign unique node IDs

PR #183, opened 2022-03-24
Add breakpoints to fig 6

awohns:
Here's what I've done so far with visualizing breakpoints in Figure 6:
inference

Below each of the inferred ARGs, I've stacked all the edges (sorted by edge start position and then edge end position), and added dashed vertical lines to indicate where breakpoints are. This shows how the ARGs change as you go along, and highlights some interesting similarities and differences between the outputs (in particular, what a JBOT actually is).

Here's another version I was experimenting with, where all the edges are labeled, and the breakpoints are also labeled below. I haven't shown this with Figure 6 because there are too many edges to actually label them.
image

And here's a simpler version of breakpoint visualization: simple, alternating black and grey rectangles. Here, the numbers indicate the number of edge diffs between marginal trees (which the earlier visualization makes clear without needing to actually put a number):
image

 Curious to hear what people think!

### DIFF
--- a/argutils/viz.py
+++ b/argutils/viz.py
@@ -49,6 +49,7 @@ def draw(
     max_edge_width=5,
     draw_edge_alpha=False,
     node_arity_colors=False,
+    edge_colors=None,
     nonsample_node_shrink=None,
     rotated_sample_labels=None,
     node_size=None,
@@ -79,13 +80,13 @@ def draw(
 
     If pos is passed in, it should be a dictionary mapping nodes to
     positions.
-    
+
     If draw_edge_widths is True, draw the widths of edges in the graphical
     representation of the tree sequence in proportion to their genomic span.
     max_edge_width specifies the maximum edge width to draw (default is 5).
     If draw_edge_widths is False, all edges are drawn with a width of
     max_edge_width.
-    
+
     If draw_edge_alpha is True, draw edges with an alpha value equal to their
     genomic span / the total sequence length of the tree sequence.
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I *really* like the first one, we should have a version of this for the paper. Is there some colour information we could include to match the edges in the two representations together?

Others also look really useful, but perhaps not in this particular context. (Sorry, short on time!)

jeromekelleher:
First example is brilliant for getting across the JBOT property

hyanwong:
P.s. @awohns you might need to rebase, as I've just merged a few minor changes to `illustrations.py`

awohns:
Here's what it looks like when ordered by the age of the child of each edge, and coloring each edge by the age of its child node (setting the color scale separately for each inferred ARG).

I think the coloring doesn't really work because the node times are over the place in each inferred ARG and we haven't reflected that in the current figure, so I don't think it's right to try to do so here. Note how many nodes are estimated to be quite old in ARGWeaver for instance.

We could try coloring by average child node depth instead?

Screen Shot 2022-03-26 at 6 23 29 PM

awohns:
How about this? Same edge ordering (by parent id) and a continuous colormap (viridis) connecting the two representations of the edges.
This was the nicest of the [perceptually uniform sequential colormaps](https://matplotlib.org/3.5.1/tutorials/colors/colormaps.html), but it does collide with the node colors @hyanwong chose.

Screen Shot 2022-03-29 at 2 07 00 PM

hyanwong:
NB probably worth rebasing your branch @awohns as some of the plot layouts have changed.

awohns:
Ok, cleaned this up and rebased: now edges are properly sorted by the time of the child node (I previously had an inconsistency between the coloring of the tree sequence and the edge representation beneath) and coloured by the [dense rank ordered child times](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rankdata.html). I've preserved your node colouring @hyanwong, and I think it will take minimal work to merge this with #191 as well.

Let me know if anything should be changed!
Screen Shot 2022-03-31 at 6 43 18 PM

PR #189, opened 2022-03-27
Use tsdate and hand-align Fig 6 nodes

### LINKED ISSUE #185: Time scales for Fig 6 plots
It is unclear what the meaning of the Y axis is in Fig 6. In fact, at the moment it is arbitrary, chosen to give the clearest layout (KwARG and ARGweaver use the default graphviz.dot Y positioning, whereas Relate and tsinfer use the ranked node times).

We have 3 options for each plot: the graphviz dot layout, the ranked times, and the true ts times. Here's what they look like (ARGweaver has  *weird* true times: presumably because of discretisation: I'm not sure logging will help here)

-----------
## Default "dot" Y positions:
Screenshot 2022-03-24 at 11 58 38

-----------
## Ranked node times:
Screenshot 2022-03-24 at 11 57 35

-----------

## True node times

Screenshot 2022-03-24 at 12 03 57

hyanwong:
All layout can then use the same Y axis (dot layout). May fix #185, if we are happy with using the dot layout as default.


jeromekelleher:
Is the idea to use tsdate on KwARG also then? That's not dated either, IIUC.

hyanwong:
I only used tsdate to put the nodes in roughly the right order for graphviz to align them. Otherwise multiple internal nodes end up being at the same level (because of shared frequencies).

The actual date values aren't shown in the plot. So I don't think we need to worry about it. It's simply a layout thing.

jeromekelleher:
I see, I thought you were talking about plotting by time. LGTM

hyanwong:
I'll merge then. Thanks for the feedback @jeromekelleher

PR #191, opened 2022-03-27
Show Relate non-JBOT

### LINKED ISSUE #161: Add Relate output to fig 6

hyanwong:
Fixes #161 ! Here's what it looks like (I haven't hand tweaked the "Relate default"node x positions, though)

Screenshot 2022-03-28 at 08 10 24


jeromekelleher:
These are basically the same I think aren't they? I guess the thing to do is to just show the "default" version as that gets across the JBOTyness of Relate trees, but that there is *some* node sharing between them?

Anyway, happy to merge whenever.

hyanwong:
> These are basically the same I think aren't they? I guess the thing to do is to just show the "default" version as that gets across the JBOTyness of Relate trees, but that there is _some_ node sharing between them?

Sure. I messed up a bit, so that the correct version actually looks like this:
Screenshot 2022-04-01 at 16 33 38

jeromekelleher:
I think we can just leave out the "Relate JBOT" column, and apply the same(ish) x tweaks to the other Relate version. Including the two will just be confusing I think.

hyanwong:
> I think we can just leave out the "Relate JBOT" column, and apply the same(ish) x tweaks to the other Relate version. Including the two will just be confusing I think.

Yes, I tend to agree, but it was worth having the JBOT in there to see what it looked like for ourselves. I think we drop it from the final version.

hyanwong:
Here's how it looks with just the "normal" relate ARG, tweaked. I guess I can simply merge this?
Screenshot 2022-04-01 at 20 28 11

jeromekelleher:
Very good, let's merge. I think it's good we're showing that Relate doesn't infer a JBOT topologically, and is actually doing something like an approximation of Espalier's MAF approach (without the explicit recombination nodes). There's much more subtlety here behind the scenes that is apparent from discussions in other papers.

jeromekelleher:
The Wa, Af and Fr trio is a great example. All four methods group them together along the whole genome. Nice illustration of tsinfer's polytomy strategy, since there's disagreement on how to group them into a binary tree.

hyanwong:
Merged. I think this means all the figures are complete now, right?

jeromekelleher:
First pass anyway, :tada:

hyanwong:
Yep - there are issues open for various improvements: I guess we can open more as we see fit.

Issue #182, opened 2022-03-23
Full sibs probably allowed in eARG encoding?


hyanwong:
In the caption for Figure 5 we say 

> Note the complex full-sib relationship between B and C; this could not be directly represented in a Griffiths-like EARG encoding (see section XXX)

But actually I think we *can* represent a full sib relationship in an eARG, can't we? It looks basically the same, like this (breakpoints are x and y)
<img width="360" alt="Screenshot 2022-03-23 at 11 21 16" src="https://user-images.githubusercontent.com/4699014/159688313-39f39af1-0302-446b-8244-f0307ea436d4.png">

It's true that we can't model this in the CwR, because the 2 recombination events and the two coalescent events are simultaneous. And it's also true that you would probably want to model the generations as gamete-to-gamete like Hudson does, and therefore place the CA events in a previous generation to the RE events. But the eARG *encoding* does allow this, I think?

jeromekelleher:
Yes, you're right I think. You'd have sample node -> RE node + breakpoint -> [parent 1, parent 2]

Issue #166, opened 2022-03-11
Rotate branches in fig 5 tree sequences


hyanwong:
The local trees in fig 5 start with their tips in the order B A C D, rather than A, B, C, D as in the graph viz. It should be easy to plot the trees with the branches rotated. I tried to change this [here](https://github.com/tskit-dev/what-is-an-arg-paper/blob/4c6f264e07cad6b7607b2e7888de0ab23a07045c/illustrations.py#L241), by changing the node order, but couldn't get it to work.

jeromekelleher:
You'd probably need to think about the order the edges are listed in, and then tell draw_svg() to use the tree order rather than the default minlex

PR #196, opened 2022-04-05
Better example for fig 5

### LINKED ISSUE #193: Add a CA-non-coalescent node to figure 5
We probably want to discuss the fact that when going from fig 5b to fig 5c we also delete any nodes that are CA only but are not associated with any coalescence (i.e. the coalescence is in non ancestral material). There are no such nodes in the current figure. I've found that the easiest way to get nice diagrams is to repeatedly simulate with different seeds. Changing the results here would probably be a bit of a faff, but worth it, I guess?

hyanwong:
And updated caption. It now looks like this, which I think is better, and has fewer trees.

Screenshot 2022-04-05 at 15 24 55

Also fixes #193 and  #166 and #182

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1077,7 +1077,7 @@ help to visualise the relative importance of different transmission pathways.
 }
 \end{figure}
 
-Given a GARG $E$ and the set of samples $S$ we can \emph{resolve} its
+Given a gARG $E$ and the set of samples $S$ we can \emph{resolve} its
 ancestry by noting that the resolved interval $Ir$ on an edge $(c,p,I)$
 is given by [Rough notes]
 
@@ -1087,7 +1087,7 @@ all $R_{u, v}$ for all the children $v$ of $u$. (this isn't quite right)
 \item The resolved interval is then the intersection of $A_u$ and the
 original inheritance interval $I_v$ on the edge.
 \item It would be good of if differentiate between the inheritance intervals
- from the GARG and the ancestry intervals of the resolved GARG.
+ from the gARG and the ancestry intervals of the resolved gARG.
 \end{itemize}
 
 This process is illustrated in Fig~\ref{fig-ancestry-resolution}.
@@ -1099,7 +1099,7 @@ genomes of the samples, there are intrinsic limits on what can
 be known about more ancient nodes, determined by the graph topology
 and the flow ancestral material through it.
 \emph{Simplification}~\citep{kelleher2018efficient} is the process
-of removing nodes and re-writing edges in a GARG to remove
+of removing nodes and re-writing edges in a gARG to remove
 various types of redundancy, while maintaining the crucial properties
 that the meaningful topology in the local trees is not affected,
 and the relationship between nodes in adjacent trees is maintained.
@@ -1111,11 +1111,11 @@ and the relationship between nodes in adjacent trees is maintained.
 \caption{\label{fig-simplification}
 ARG simplification. In all cases the graph visualization is shown on the left
 and the equivalent marginal trees on the right.
-(A) An ARG simulated from a diploid Wright-Fisher
-model. Note that the genome \textsf{K} has three children and the genomes
+(A) A gARG simulated from a diploid Wright-Fisher
+model. Note that the genome \textsf{K} has three children, and the genomes
 \textsf{E} and \textsf{G} simultaneously have multiple parents and multiple children;
-neither could not be directly represented in a Griffiths-like EARG encoding
-(see section XXX).
+neither could be directly represented in a Griffiths-like eARG encoding
+(see section XXX). The penultimate local tree is highlighted to aid discussion (see section YYY).
 (B) Simplified to remove all
 % trying out this terminology - can define in the text
 1-connected graph components (e.g., diamonds such as \textsf{JLMN}).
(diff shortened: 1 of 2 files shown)


hyanwong:
The CA/non-coal node here is R, removed when going from b->c (along with node N which is not a CA node)

hyanwong:
There's an interesting observation here:

**Fig a**: Since each breakpoint is at a different position, the
the number of recombinant nodes is equal to the number of breakpoints (which is
equal to the number of local trees plus one)

**Fig b**: As in A, the number of breakpoints is equal to the number of recombinant nodes

**Fig c**: Recombinants may have more than 2 parents representing multiple stacked recombinations (e.g. node
`E`. Each additional parent implies a new breakpoint, so although there are only
3 recombinant nodes, node `E` accounts for 2 breakpoints, and hence there are 4 breakpoints (and 5 local trees).

**Fig d**: There is no direct link between the number of node parents and the number of trees.

Is that right? If we have all the coalescent nodes with their unary parts, can we calculate the number of trees like this? I suspect the logic is more complicated, but possibly worth figuring out.

hyanwong:
If we like this, we'll need to change the text discussion, which talks about specific nodes in the viz.

jeromekelleher:
Looks great. I don't have time to think hard about the number of recombination nodes/distinct trees, but it sounds plausible. Would need to note that this is "infinite sites" recombination somewhere, though. I guess that's what you're saying the first sentence of (a) above?

jeromekelleher (approved):

hyanwong:
> Looks great.

Good. I think we should merge then (I could change the text discussion first).

> I don't have time to think hard about the number of recombination nodes/distinct trees, but it sounds plausible. 

Yes, this is a TODO.

> Would need to note that this is "infinite sites" recombination somewhere, though. I guess that's what you're saying the first sentence of (a) above?

Exactly.

hyanwong:
I've changed the paper text, and to help with the discussion, highlighted one of the local trees in yellow. What do you think @jeromekelleher ?

Screenshot 2022-04-06 at 11 23 40

paper.tex
```diff
@@ -1109,84 +1109,110 @@ \section*{ARG simplification}
...
-ARG simplification. (A) An ARG simulated from a diploid Wright-Fisher
-model. Note the complex full-sib relationship between B and C; this
-could not be directly represented in a Griffiths-like EARG encoding
-(see section XXX).
+ARG simplification. In all cases the graph visualization is shown on the left
+and the equivalent marginal trees on the right.
+(A) A gARG simulated from a diploid Wright-Fisher
+model. Note that the genome \textsf{K} has three children, and the genomes
+\textsf{E} and \textsf{G} simultaneously have multiple parents and multiple children;
+neither could not be directly represented in a Griffiths-like eARG encoding
```
jeromekelleher, line 1117:
```suggestion
neither could be directly represented in a Griffiths-like eARG encoding
```

paper.tex
```diff
@@ -1109,84 +1109,110 @@ \section*{ARG simplification}
...
-(E) The local trees for ARG (A).
-(F) The local trees for ARG (D).
+1-connected graph components (e.g., diamonds such as \textsf{JLMN}).
+(C) Remove nodes that are unary in all local trees. This removes both recombinant nodes
+that have only one child (e.g. \textsf{N}) and nodes that have more
+than one child (i.e. are ``common ancestors'') but never represent coalescences
+in the local trees (e.g. \textsf{R}).
+(D) Rewrite edges to bypass any unary nodes in local trees. Note that this loses lineage
+information such as the fact that \textsf{K} is connected to \textsf{A} via \textsf{E}; it
+can also create many extra roots.
```
jeromekelleher, line 1128: Not sure about this last bit - people might be confused between roots in the trees and roots of the graph. I think probably just delete after the semicolon.

paper.tex
```diff
@@ -1109,84 +1109,110 @@ \section*{ARG simplification}
...
-local trees along the genome for a simulated ARG
-(Fig.~\ref{fig-simplification}A). Consider the right-most tree,
-and the path above node $D$. Before $D$ meets its common ancestor
-over the interval $(93, 100]$ with the other samples it passes through
-nodes $H, K, N, R$ and $S$, reflecting the corresponding
-traversal through the ARG (see section XXX). Such unary nodes
-in a local tree are not usually considered significant,
+one child. Fig.~\ref{fig-simplification} show a series of successive steps of
+simplification, starting with a complete gARG simulated under a Wright-Fisher
+model, shown both as a mathematical graph on the left, and the equivalent set of 6
```
jeromekelleher, line 1142:
```suggestion
model, shown both as a  graph on the left, and the equivalent set of 6
```
No reader who has got to this point needs the "mathematical" (and including it would probably confuse them)

paper.tex
```diff
@@ -1109,84 +1109,110 @@ \section*{ARG simplification}
...
 detectable by information from the tree. % weak
-The nodes in a tree usually mark the points where samples find a
-\emph{common} ancestor, rather than focusing on specific
-ancestors. For most purposes, we would consider an edge
-directly joining sample $D$ to the node $T$ to be more
-useful and informative. Simplification is the process of removing
-these uninformative and (for most computational purposes)
-inefficient unary paths.
+In the local trees, attention is usually focussed upon points where upwards
+lineages meet in a \emph{common} ancestor (the MRCA node, or coalescence point).
```
jeromekelleher, line 1153: I think the original first sentence is better here. The "coalescence point" terminology doesn't do much for me, and I don't see what talking about upwards lineages helps.

paper.tex
```diff
@@ -1109,84 +1109,110 @@ \section*{ARG simplification}
...
+(node \textsf{R} in Fig.~\ref{fig-simplification}B).
+% point out that this is why Hudson refers to CA nodes, not coalescence nodes
+Such nodes are not singly connected in the graph, but are nevertheless unary in
+all of the local trees in Fig.~\ref{fig-simplification}B. The operation to remove them,
+the results of which are shown in Fig.~\ref{fig-simplification}C,
+therefore requires knowledge not just of the graph topology, but also of the
+edge annotations. Removal of recombinant nodes can produce graph nodes with
+more than 2 parents (e.g. node \textsf{E}); likewise, removal of
+common ancestor but non-coalescent nodes can produce graph nodes with
+more than 2 children (e.g. node \textsf{S}). These representing a ``stacking up'' of multiple
```
jeromekelleher, line 1194: Last sentence of para doesn't parse I think.

paper.tex
```diff
@@ -1109,84 +1109,110 @@ \section*{ARG simplification}
...
-The final level of simplification is to remove \emph{all}
-unary nodes from the local trees (Fig.~\ref{fig-simplification}D
-and E). [And continue]
+local tree. For example, node  \textsf{K} is unary in the second tree
+of Fig.~\ref{fig-simplification}C, but is either binary
+or ternary in all subsequent trees (recall this is a Wright-Fisher
+simulation). The final level of simplification is to alter the edge annotations
+such that, although no nodes are removed from the graph, \emph{all}
+unary nodes disappear from the local trees (Fig.~\ref{fig-simplification}D).
+It is common for this operation to remove nodes above the top MRCA node
```
jeromekelleher, line 1208: I'd rephrase this a bit, but can wait for another round of revisions.

jeromekelleher (approved):
Looks good, some minor comments.

paper.tex
hyanwong: Well spotted, thanks.

paper.tex
hyanwong: OK, but I think "uninformative" is a little strong. I put in coalescence, as that's what most popgen people would use to describe such a node in the local trees, so thought it would aid familiarity. Similarly with MRCA. How about:

> The nodes in a tree usually mark the points where samples find a \emph{common} ancestor, such as node $Q$ in the highlighted tree. For many purposes, we would therefore consider a path directly joining sample $A$ to node $Q$ to be of equivalent status to the path containing the intermediate unary nodes. Simplification is the process of removing these less informative and (for most computational purposes) inefficient nodes from the local trees and the corresponding annotated graph.

I think we need to talk about removing nodes from the trees because the last step (c->d) doesn't remove graph nodes.

paper.tex
hyanwong: Yeah - I put in the bit about roots with the aim of rephrasing later, but though it worth adding in case we did want to say something.

hyanwong:
> Looks good, some minor comments.

Thanks. Helpful comments.

paper.tex
JereKoskela: Very minor comment which may not end up being relevant here depending on how things are drafted, but I'm always a bit hesitant to call things "upward" or "downward". We're very consistent with drawing all of our trees in the same direction, but there is nothing close to a universal convention in the literature. I'd prefer "leafward" and "rootward", or words to that effect.

hyanwong: Yes, ISWYM. It was to avoid "backward" vs "forward", because I think some people instinctively think backwards anyway, so using that adjective can be a bit jarring.

Worth bearing in mind in general, I agree. Thanks

jeromekelleher:
LGTM, merge away whenever you're happy

PR #201, opened 2022-04-06
Experiment with using named labels for section refs

hyanwong:


hyanwong:
Merging: uncontroversial I think, and good to have example code showing how to cross-ref.

Issue #198, opened 2022-04-06
Content for Ancestral graphs and stochastic processes section


jeromekelleher:
- Move background content for phylogenetic networks to the end of this section (+ condense)
- Move the figure 1 and technical text around hudson's alg to the Ancestry Resolution section

This should make it a relatively lightweight background section, describing where the ARG came from, and what the related concepts are (which we're not going to talk about again).

jeromekelleher:
In terms of where we move Fig 1 (not the content of the sectino in question):

Arguably the current figure 1 is better associated with the sectin on Sampling Probabilities. This gives us that opportuntity to contrast the Big ARG's likelihood (simple) vs the little ARG.

PR #202, opened 2022-04-06
Move Hudson discussion into later section

hyanwong:
Along with Fig 1 and the phylo network discussion.

WIP. This is the big restructing as discussed in #198. Needs the section on *Ancestry resolution* revising to flow better.


hyanwong:
Note that it would be useful to use what is now Fig 3b to highlight the parts of the graph corresponding to HUD (or the nodes/paths which are no longer necessary) . Perhaps something like this:

image

jeromekelleher:
Restructure looks good to me. I say we merge and discuss other changes separately.

PR #218, opened 2022-07-06
Add a markdown summary

hyanwong:
For discussion only


jeromekelleher:
Things have moved on again since this was done @hyanwong - shall we close, or merge and keep as a record for future reference?

hyanwong:
I think it's useful. Let's leave open as a PR, and I can add commits to deal with the new structure, then we can merge.

Issue #199, opened 2022-04-06
ARG Inference methods section revision


jeromekelleher:
~~1. Remove the subsections
2. Condense things down to a few paragraphs.~~

~~No need to name all the methods, we just want a rough classification following existing structure. Basically one paragraph discussing the parsimony and hueristics, and one about model based methods. (Perhaps better to put the model sampling first, for clarity).~~ (now done)

Then some paragraphs discussing Fig 6. Things to say

1. They're all ARGs. 
2. Point out the "recombinant" nodes in Relate and tsinfer. Explain how Relate and tsinfer make ARG-like things in more detail.
4. In contrast kwarg and ARGweaver have exact recombination nodes.
5. We can see commonalities in these inferred structures, but there are also substantial differences. We have some quantitative information about the ARGs, which may be helpful for later comparison.
6. They're all ARGs and they're all represented using tskit succinct tree sequences here.

a-ignatieva:
I will!

PR #226, opened 2022-07-11
Figure 6 narrative

a-ignatieva:
First draft re #199


a-ignatieva:
- Is it clear enough that tsinfer and Relate produce ARGs here? I could say something about "shared edges" if the "correlated local trees without explicit SPRs" bit isn't doing the job
- I think the ARGs might look more similar if we made an effort to equalise how happy they are to add more mutations (through setting event costs for KwARG, and mutation/recombination rates in ARGweaver and Relate, and the mismatch ratio in tsinfer). But this might be unnecessary for the purposes of this paragraph, so I've just said that this is a source of differences.

JereKoskela:
I like how this reads. I've made a minor in-line comment re: your first bullet point. For the other two:

I think we should stick to defaults as much as possible. That's what people will see if they try similar things themselves. We should say that we're doing that, but I think what you've written covers it.

Referring back to Fig 4 is a nice idea. I'm in favour.

paper.tex
```diff
@@ -1422,6 +1424,14 @@ \section*{ARG inference methods}\label{ARG_inference_methods}
...
+
+\texttt{KwARG} \citep{ignatieva2021kwarg} produces a parsimonious ARG with seven recombination breakpoints, which has been shown to be the true minimum required to explain the dataset when disallowing recurrent mutation \citep{song2003parsimonious}. An ARG drawn from the posterior sample produced by \texttt{ARGweaver} has 17 breakpoints. Both tools explicitly record the sequence undergoing each recombination event, and the precise location of the breakpoint along the genome. \texttt{Relate} \citep{speidel2019method} and \texttt{tsinfer} \citep{kelleher2019inferring} instead construct a sequence of correlated local trees, but without specifying the precise positions of recombination events (i.e.\ reconstructing the local trees without explicitly specifying the SPR operations required to transition between them). The number of inferred trees also depends on the parameter settings; all four tools can allow for the presence of sequencing errors and recurrent mutations, reducing the number of inferred local trees (in this case, based on the input mutation and recombination rate \texttt{Relate} hence posits only two recombination breakpoints).
```
hyanwong, line 1431: I worry that saying 

> "Relate" and "tsinfer" instead construct a sequence of correlated local trees

will reinforce the (mistaken) assumption that these are both JBOTs. I would quite like to rephrase this?

hyanwong: How about

> \texttt{Relate} \citep{speidel2019method} and \texttt{tsinfer} \citep{kelleher2019inferring} do not place recombination events onto the ARG topology, so do not encode the transitions between local trees in terms of explicit SPR operations.

hyanwong (commented):
Looks good generally, but I would be careful about saying that relate & tsinfer are simply a set of local trees, even if the word "correlated" is in there. This has been a source of considerable misunderstanding in the past, and one of the main points of the paper is to say that they both do produce ARGs.

paper.tex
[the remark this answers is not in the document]
a-ignatieva: Have added a comment to this effect:
> \texttt{KwARG} \citep{ignatieva2021kwarg} produces a parsimonious ARG with seven recombination breakpoints, which has been shown to be the true minimum required to explain the dataset when disallowing recurrent mutation \citep{song2003parsimonious}; this is a lower bound that, in general, significantly underestimates the actual number of recombinations that might have occurred.

a-ignatieva: Cheers, I think this is more to the point! 
Regarding Relate, I've said
> \texttt{Relate} first infers local tree topologies and then identifies edges that persist from one tree to the next as a separate step, while \texttt{tsinfer} explicitly constructs nodes that are shared across local trees.
I think in general Relate is pretty good at identifying equivalent branches.

paper.tex
```diff
@@ -1422,6 +1424,14 @@ \section*{ARG inference methods}\label{ARG_inference_methods}
...
+
+There is some consistency among the four inferred ARGs (such as the samples FR-F, WA-F and Af-F being grouped into the same clade), but also considerable differences in the order and number of events. Note that all four ARGs can be represented in the succinct tree sequence format, with the ARGs produced by \texttt{KwARG} and \texttt{ARGweaver} containing more nodes to record the recombination events (corresponding to Figure 4 stage `B'), and the ARGs output by \texttt{tsinfer} and \texttt{Relate} being, in essence, simplified (Figure 4 stage `D').
```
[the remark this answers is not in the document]
a-ignatieva: Can we get away with saying something like
> The ARGs produced by \texttt{KwARG} and \texttt{ARGweaver} contain more nodes to record the recombination events (corresponding to Figure 4 stage `B'), while the ARGs output by \texttt{tsinfer} and \texttt{Relate} being, in essence, simplified (roughly corresponding to Figure 4 stages `C' and 'D', respectively).

paper.tex
hyanwong: Yes, that sounds good to me. I do think in the figure that the Relate ARG looks rather different from the others, rather like a series of trees, so maybe worth commenting on somewhere else?

hyanwong:
LGTM. One very minor comment:

> therefore the transitions between local trees (and corresponding SPR operations) are not explicitly inferred

is it true that "the transitions between local trees are not explicitly inferred" in both cases (which is how I read it with the brackets). I think that both tsinfer and relate do attempt to infer recombination breakpoints along the genome, but they don't attempt to place the nodes on the topology. So I might say "the transitions between local trees are not explicitly inferred in terms of SPR operations"

hyanwong:
LGTM. merging

PR #227, opened 2022-07-13
Rephrase "Ancestry Resolution"

### LINKED ISSUE #212: Ancestry resolution section outline
With https://github.com/tskit-dev/what-is-an-arg-paper/pull/202 we have put the section describing the Hudson algorithm into the section entitled "Ancestry  Resolution". It is also possible that we should move the details of the calculated probabilities to accompany figure 5 (the one that used to be figure 1) in the "Evaluation of sampling probabilities" section. Either way, the text in the "Ancestry Resolution" section needs finessing to make sense. Here's a suggested flow:

* Up to this point in the paper we have shown edge annotations that describe "immediate inheritance"
* ARGs are inherently retrospective, defined WRT samples
* We can define a sample resolved ARG by trimming edges and removing nodes that are non-ancestral
* The most efficient way to do this is to follow the same logic as in the Hudson algorithm (and as illustrated in Fig 3B, the Wiuf & Hein one)
* We call this algorithm, which is like running Hudson on a fixed topology, "XXX-simplification" (see #210), and it is an integral part of using tskit when creating ARGs from forwards simulations.
* ~~We outline the XXX-simplification algorithm and compare it to the Hudson algorithm (but I suggest that we move the formal description to supplementary text, see #213)~~
* As a result of simplification, non-ancestral nodes can get removed (see Fig 3B) and optionally we can also remove nodes above the local roots (need to highlight these in Fig 3B, see e.g. https://github.com/tskit-dev/what-is-an-arg-paper/pull/202#issuecomment-1090460472)
* The reduction in number of nodes require to trace means that the "simplified" structure (or little ARG) is much smaller than the full Griffiths gARG encoding (a.k.a. big ARG). 

Then we could either go on to the "ARG simplification" section (as we do at present) or move the "sampling probabilities" section next instead.

hyanwong:
I've been ruthless here and commented a lot of technical stuff out, in an attempt to make this section both shorter and more accessible. We can uncomment the paragraphs that we think are needed in the main text. All feedback welcome.

When completed, is intended to fix #212

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -615,7 +615,7 @@ required if, for example, we have a deeply sequenced pedigree as in \cite{RosFre
 % e.g. Honeybees? Be great to cite an example!
 % Note by Yan - if we actually have the sequence, then this turns into
 % a sample node, so would be normally included anyway. Perhaps, therefore
-% we should talk about these "pass though" nodes in the context
+% we should talk about these ``pass though'' nodes in the context
 % of internal samples.
 % Gregor: There are no good deeply sequenced pedigrees in honeybees, but sex
 % chromosome behaves like haplodiploid system, also any deep pedigree works here, so I cited our pig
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -724,77 +724,71 @@ \section*{Genome ARGs}
...
-disjoint genomic intervals $(\ell, r]$
-over which genome $c$ inherits from $p$, and there is a path from
-$c$ to some non-empty subset of $S$ through the local trees.
-% Should probably define how we build the local trees somewhere.
-
+The edge annotations we have described in Fig 1 and 2 show the regions of genome
+that have been passed from a parent to its child in their entirety. In effect, all edges
+span the entire length of the genome, apart from those that connect two parents
+to a child in which they recombine. In this form of edge annotation, which we call
+``entire edge encoding'', it is inefficient to extract the local trees from the ARG structure.
```
jeromekelleher, line 735: This phrase doesn't do anything for me - I think it'll just confuse things more. What's "entire" about it? I think "local" or "immediate" inheritance is much clearer.

paper.tex
```diff
@@ -724,77 +724,71 @@ \section*{Genome ARGs}
...
+been inherited by the samples. This also allows the removal of nodes which are not
+genetically ancestral to the samples.
 
 \begin{figure}
 \centering
 \includegraphics[width=\textwidth]{illustrations/ancestry-resolution}
 \caption{\label{fig-ancestry-resolution}
 The \citet[][fig. 1]{wiuf1999recombination} example graph as a gARG, with contrasting
 edge annotation schemes.
-(A) Edge annotations that reflect immediate inheritance, i.e. transmission of
```
jeromekelleher, line 763: I think these changes are worse actually "encoding" is a term we want to avoid where we can. We're recording different information here, not using a different way to write down the same information.

jeromekelleher (commented):
I don't like the "entire" vs "sample resolved encoding" terminology here, and would change it back to something closer to what it is now. However, it's probably simplest to merge this and revise, rather than go back and forth on the PR

paper.tex
hyanwong: Good point. Let's discuss in our meeting later today.

paper.tex
hyanwong: Sure, let's pick something that's the least confusing. We wanted to avoid "local" though. We can switch back to immediate (which I think is what we had before).

hyanwong: As discussed online, changing to "immediate edge annotation(s)", with a comment about better names

hyanwong:
Merged after discussion with @jeromekelleher. Material can be reinstated by uncommenting.

PR #229, opened 2022-07-14
Sampling probabilities rewrite

### LINKED ISSUE #223: Integrate/rework the “sampling probabilities” and “inference methods” section
As discussed @JereKoskela has ideas of how to do this.

JereKoskela:
A first pass at shortening the sampling probability description and merging it with the inference methods section. I've left in the notation because I thought the description became more confusing without it, but pretty much all actual detail has been cut. Comments very welcome.

Closes #223

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -1090,9 +1090,6 @@ leaf-labelled binary trees.
 % Clearly, we can only ever hope to infer the fraction of an ancestor's
 % genome that is ancestral to the sample.
 
-\section*{Evaluation of sampling probabilities}\label{probabilities}
-
-
 \begin{figure}[ht]
 \centering
 % FIXME this is a quick nasty hack to make the figure a bit smaller and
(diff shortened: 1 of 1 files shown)


hyanwong:
I like this. I wonder if we can keep a phrase in about it being necessary to have a sample-resolved gARG to do this?

I assume some of the removed material will go in the appendix?

JereKoskela:
> I like this. I wonder if we can keep a phrase in about it being necessary to have a sample-resolved gARG to do this?

I left it out because just having a sample-resolved gARG isn't enough: we'd still need to describe the rules for grouping edges into lineages, which seemed like the sort of technical detail we wanted to cut. I thought it better to remove the whole thing rather than give a half-explanation.

> I assume some of the removed material will go in the appendix?

I think this might be a good place to bring back the gARG sentence, since we can afford to give a complete description. I could briefly refer to the Appendix in the main text to link the two.

jeromekelleher (approved):
LGTM

hyanwong:
Do you want this merged now @JereKoskela, or shall I wait until the appendix stuff is in?

hyanwong:
Fine, merging.

PR #231, opened 2022-07-19
Args as stochastic processes removal

hyanwong:
I have removed the stochastic processes section by moving some parts to the discussion and commenting out some of the parts that were not moved. Please do read through the intro and the first 4 paras of the "Event ARG data structure" section to see if it still flows and makes sense.

The idea here is to get quickly to the eARG stuff without distracting the reader too much with the history (although we do need a little of that).


hyanwong:
Great, thanks for looking Jerome. Merging

Issue #221, opened 2022-07-07
Appendix with Fig 5


hyanwong:
Create a new “Big ARG vs Little ARG” appendix containing Fig 5 and remaining material from the "ARGs as Stochastic Processes" section.

JereKoskela:
I'll close this in light of #240. We can open a more detailed issue if there are tweaks to be made.

PR #237, opened 2022-07-31
Add "history" section & outline Big and Little ARG appendix

jeromekelleher:
We need to explain the history of where the terms came from, and how the ARG-as-a-process vs ARG-as-a-data-structure confusion came about. This is a first pass at that, mostly pulling together pre-existing content.

Needs quite a lot of polish, but I think the basic idea is right.

Also Partially addresses #221 

Most of the changes towards the end of the diff are just noise - I moved around a lot of content and Git has got confused.


paper.tex
```diff
@@ -126,110 +130,134 @@ \section*{Introduction}
...
+sampling methods to  exploit.
+The focus subsequently shifted to the more
+tractable---but still NP-hard~\citep{wang2001perfect}---problem of computing
+the minimum number of recombinations required
+to explain an input dataset~\citep{myers2003bounds}, and inferring the corresponding
+ARG realisations~\citep{song2003parsimonious,song2005efficient,lyngso2005minimum}.
+The previously close correspondance
+between the ARG as a mathematical representation of the coalescent process
+and as concrete realisations of ancestry was now broken,
+since such parsimonious ARGs are extremely unlikely under the
```
hyanwong, line 200: I'm not sure it's the value of the likelihood of a parsimonious ARG that is important, it't the fact that it is not constructed WRT to the likelihood itself.

hyanwong (approved):
Great work @jeromekelleher. I like this and think we should merge it and iterate.

paper.tex
jeromekelleher: Right, but we were now saying that we were interested in ARGs that would be considered neglibly likely under the model, and therefore that the model itself wasn't the thing of primary interest. Maybe this isn't getting across?

a-ignatieva: Perhaps the disctinction to emphasise should just be that the parsimony-based approach ignores branch lengths and focussed on sequences of events / topology? Leading to the concept of the ARG as a data structure.

jeromekelleher: Good point, I forgot about the lack of branch lenghts! That's definitely something to mention. I'll rejig a bit and drop some notes tomorrow.

jeromekelleher:
OK, merging this. Will need significant follow-up work, but I'm happy that the overall structure and narrative works for this section at least.

PR #238, opened 2022-08-01
Draft of eARG section

jeromekelleher:
Here's an update to the eARG section. The idea is to avoid the complications of node vs edge annotations entirely here. We call the thing that we're contrasting with an eARG, try to be concrete about what it means, and then point out some of its shortcomings. We can discuss node-vs-edge annotations (and point out that this is actually partially orthogonal to event-vs-genome ARGs) later.

Will need some revision, but hopefully the rough outline of the content is there.


jeromekelleher:
I'm going to ahead and merge this, as there's no objections.

PR #240, opened 2022-08-03
ARG process appendix

JereKoskela:
A first draft of the appendix. I also added reference sentences to it in the main text when we bring up either stochastic processes or sampling probabilities, and slightly shuffled things in Figure 6 to facilitate making it a bit bigger.

I've used the terms "big ARG process" and "little ARG process" to talk about the Griffiths and Hudson processes, to avoid adding to the confusion about using just "ARG" for both the data structure and for a stochastic process. It didn't feel right to call them the "big" and "little coalescent with recombination", because that gets quite cumbersome.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -190,6 +190,7 @@ requires exponential time to simulate
 to describe recombinant ancestry as single object that
 could be inferred in a way
 that was missing from Hudson's treatments.
+An overview of the Hudson and Griffiths stochastic processes is provided in the appendix.
 
 Early work on ARG inference focused on the problem of
 inferring parameters of the
@@ -1155,6 +1156,7 @@ number of links available for recombination (i.e.\ ones at which a recombination
 would split ancestral material) at each time \citep[Eq.\ (3)]{mahmoudi2022bayesian}.
 These require a data structure from which shared edges on different trees are easy to identify,
 and which encodes recombination events and times explicitly, e.g.\ using nodes.
+We will provide a more detailed discussion on sampling probability evaluation in the appendix.
 
 \subsection*{Differences in reconstructed ARGs}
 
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -1291,11 +1293,82 @@ \section*{Appendix}
...
+$(\ell, r, a_1 + a_2)$, and if $a_1 + a_2 < n$ it is included in the
+ancestor lineage. Otherwise, if $a_1 + a_2 = n$, we have found
+the most recent common ancestor of all samples in the interval$[\ell, r)$
+and do not need to simulate its history any further.
+Non-overlapping intervals from the two lineages are included
+ in the ancestor lineage without changes. Eventually,
+we find resultant lineages in which all segments have fully coalesced,
+and so the number of extant lineages gradually falls to zero.
+
+In the Griffiths formulation, each edge in the graph corresponds to an extant
```
jeromekelleher, line 1332:
```suggestion
In the Griffiths formulation (the Big ARG process), each edge in the graph corresponds to an extant
```

paper.tex
```diff
@@ -1291,11 +1293,82 @@ \section*{Appendix}
...
+substantial cost, however, if we wish to use it as a practical means of
+simulating recombinant ancestries. 
+The number of events in the big ARG all the way back to the GMRCA
+is $O(e^\rho)$~\citep{griffiths1997ancestral}, whereas the number
+of events required to simulate the little ARG is
+$O(\rho^2)$~\citep{hein2004gene,baumdicker2021efficient}.
+This disparity arises because the majority of the events in the big ARG are 
+recombination events which occur outside of ancestral material,
+and this do not have any bearing
+on the ancestry of the initial sample.
```
jeromekelleher, line 1355: Maybe add another sentence here like "Because we don't keep track of the distribution of ancestral material during the process, we generate a vastly larger graph."

paper.tex
```diff
@@ -1291,11 +1293,82 @@ \section*{Appendix}
...
+
+Figure \ref{hudson_vs_bigARG} illustrates the more complex state space of the little ARG process, as well as the extra events which occur in the big ARG process.
+Moreover, it depicts the rates of common ancestors and recombination events in each interval of time of the realisations.
+In order to evaluate these rates, and hence the sampling probability (see e.g. \cite[Equation (3)]{mahmoudi2022bayesian}), it is necessary to know the number of lineages and number of extant links available for recombination in each time interval.
+Note that this information cannot be recovered from the gARG encoding depicted in e.g. Figure \ref{fig-ancestry-resolution}.
+For example, it clear that a recombination takes place between nodes \textsf{b} and \textsf{d} as well as \textsf{e}, but the exact time of the event is ambiguous, and thus so is the number of lineages during the time interval.
```
jeromekelleher, line 1361:
```suggestion
For example, it is clear that a recombination takes place between nodes \textsf{b} and \textsf{d} as well as \textsf{e}, but the exact time of the event is ambiguous, and thus so is the number of lineages during the time interval.
```

jeromekelleher (approved):
LGTM!

A few minor suggestions in the above.

We should probably cite the msprime 1 paper for further info on computing the likelihood.

Would you mind splitting long lines please? Git works much better with lines broken at < 100 chars.

paper.tex
JereKoskela: Thanks, fixed.

paper.tex
JereKoskela: Done, thanks.

paper.tex
JereKoskela: Thanks, fixed.

JereKoskela:
Thanks @jeromekelleher! I included your suggestions, added a citation to the msprime 1.0 paper, and cut all lines to under 100 characters.

PR #249, opened 2022-08-25
Precision of ancestry section

jeromekelleher:
WIP writing the section on how precise our representation of genetic ancestry can/should be.

As part of this, I'm merging in the discussion of Fig 5 as a way to illustrate. This lets us get this nice figure in, but avoid the dull descriptive text that would be required if we're just going to talk about them on their own.

Still working on it, hopefully will finish up a draft tomorrow but opening this PR so folks can see what's happening if interested.


jeromekelleher:
Merging this much - still a WIP but no point in holding up other edits.

Issue #251, opened 2022-08-26
Break Fig 1A into two tables


jeromekelleher:
I think the current layout of Fig 1A is misleading:

![Screenshot from 2022-08-26 14-42-40](https://user-images.githubusercontent.com/2664569/186917290-4d3c0ff4-38df-45e3-9470-25ab0fb3935c.png)

This looks like we have a mapping from the node to two things, the breakpoint and an edge. This is incorrect, as we actually have a mapping from node to breakpoint plus and *unrelated* ordered list of edges. We need to show that the encoding consists of these two distinct things.

Particularly with 1C in mind, where a similar layout is used to show two things that are related (edges and intervals).

hyanwong:
A good point. I agree that it should be two separate tables.

PR #260, opened 2022-09-13
Merge eARG and gARG sections

### LINKED ISSUE #256: Refactor eARG and gARG section
I think it would help the narrative if we reworked the Event ARG and Genome ARG sections a bit:

1. Change Figure 1 to just show a classical eARG (i.e., remove C). We can change the tabular representation to something simpler where we have three columns, ``node``, ``parents`` and ``breakpoint``. This would be easier to describe (and close #251). Probably most intuitive to show the graph as subfig A and the tabular representation as B. Rewrite caption so say "this is a classical eARG".
2. Remove the second paragraph "To be concrete, we can formally..". We don't really use this formal description much, and where we do we can easily rewrite in terms of standard "two parents" 
3. Finish up the section by just listing a few of the weaknesses of the encoding - we spend a lot of the rest of the paper talking about this, so we don't need much detail.

Then, we discuss the process of converting an eARG to a gARG in the "Inheritance and ancestral material" section, where we use the WH example in Figure to illustrate (#253)

Hopefully rejigging like this will bring readers of all experience levels with us, making sure that we're clear on what we mean about things, without burdening them with lots of terminology that we expect them to remember between sections.

jeromekelleher:
Closes #256 #251

In an attempt to shorten things down and make the text a bit snappier (as discussed in #256) I've merged the aARG and gARG sections.

I've been quite ruthless about cutting stuff out, but kept most of the old content commented out so we can merge bits of it back in later.

I've probably gone too far with the cutting, but how does the new outline look?

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -561,8 +561,7 @@ to describe patterns of genetic inheritance must be in terms
 of specific causal events. As we demonstrate in subsequent sections,
 there are significant advantages to using
 a slightly different formulation, which
-we call a ``genome ARG'' (gARG) to distinguish it from the classical
-eARG described above.
+we call a ``genome ARG'' (gARG).
 First, let us define a genome as the complete set of genetic material that a child
 inherits from one parent (i.e., haploids have one genome, diploids two, etc.).
 Then, a gARG is a graph in which nodes represent
@@ -600,15 +599,6 @@ whereas \textsf{b} was inherited directly from $I_4$'s genome \textsf{g} without
 recombination.
 Arbitrary patterns of genetic inheritance (gene conversions, multiple recombinations,
 etc) can be expressed without difficulty.
-
-% To distinguish from the classical Griffiths eARG,
-% we refer to this encoding of an ARG data structure
-% as a ``Genome ARG'', or gARG.
-% The distinction between nodes representing events
-% and genomes may seem pedantic at first, but as we
-% see in later sections, we gain a great deal of flexibility
-% and computational efficiency by decoupling our representation
-% of genetic ancestry from the events that generated it.
 
 \begin{figure}
 \begin{center}
@@ -720,12 +710,14 @@ there are substantial practical gains
 from using the gARG encoding, allowing differential
 levels of information about the recombination process to be
 represented in a computionally efficient way.
-The key to this is not in how we interpret nodes, but in how
+The key to this is not so much in how we interpret nodes
+(see Appendix XXX), but specifically in how
 we represent the process of genetic inheritance.
 The critical distinction is that
-the gARG encoding allows us to associate
-genetic inheritance in non-contiguous chunks of genome
-between an ancestor and a descendant.
+the gARG encoding allows us to directly describe
+genetic inheritance in multiple non-contiguous genome intervals
+between an ancestor and a descendant, something that
+cannot be done by storing recombination breakpoints.
 
 \section*{Inheritance and ancestral material}\label{Ancestry_resolution}
 The edges in a gARG represent a path of genetic inheritance from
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -634,112 +639,90 @@ \section*{Genome ARGs}
...
-classical eARG (as defined in the XXX section) with the gARG definition
-here. However, it is important to note that many of the benefits of the
+classical eARG with the gARG definition here.
+However, it is important to note that many of the benefits of the
 gARG encoding that we discuss are due to the encoding of inheritance via
 edges rather than nodes.
-Whether one regards nodes as genomes or events is orthogonal,
-and something of a philosophical question which makes little
-difference from a computational perspective.
+Whether one regards nodes as genomes or events is essentially
```
hyanwong, line 719: Perhaps a useful distinction is that edge encoding is useful for sample resolving, whereas nodes-as-genomes (coupled with edge encoding) is useful for allowing necessary imprecision? Later, we could work this up into a replacement sentence here somehow.

paper.tex
```diff
@@ -634,112 +639,90 @@ \section*{Genome ARGs}
...
+The eARG and gARG definitions provided here are apparently very
+similar, but, as we shall see, there are many advantages to decoupling
+our representation of patterns of genetic inheritance
+from the events that generated them.
 For simplicity, throughout the rest of the paper we will contrast the
-classical eARG (as defined in the XXX section) with the gARG definition
-here. However, it is important to note that many of the benefits of the
+classical eARG with the gARG definition here.
+However, it is important to note that many of the benefits of the
 gARG encoding that we discuss are due to the encoding of inheritance via
```
hyanwong, line 717: I think we wanted to reserve "encoding" for a different purpose, and refer to "annotation" of nodes vs edges, right?

hyanwong: So more like:
> However, it is important to note that many of the benefits of the gARG
encoding that we discuss are due to tracing paths of inheritance by annotating
edges with intervals rather than annotating nodes with breakpoints.

hyanwong:
The one thing that we lose with the merging of the sections is the idea that if you *do* want to describe an eARG, you can still use tskit to do so. This also ties in with the idea of using 2 recombination nodes as a representation. It might be worth thinking about if/how we can slip that in, without making too much of a fuss about it.

paper.tex
jeromekelleher: Yes, we need to get at something like this. I'll have another go at this final para.

paper.tex
```diff
@@ -707,17 +710,22 @@ \section*{Event vs Genome ARGs}\label{eARG}
...
+% Try to tackle this head on by getting to the heart of the distinction.
+The difference between eARGs and gARGs are subtle, and whether
+we choose to interpret nodes as events or genomes
+is often of little practical importance.
+However, as we show in the remainder of this article,
+there are substantial practical gains
+from using the gARG encoding, allowing differential
+levels of information about the recombination process to be
+represented in a computionally efficient way.
+The key to this is not in how we interpret nodes, but in how
```
hyanwong, line 723: "Not" is a little too strong, I think, as we do require nodes to be genomes if we are to condense multiple coalescences or recombinations into a single node (not allowed if the nodes are specific events). So perhaps something like "The key to this is less to do with the exact interpretation of nodes, and more to do with how we represent genetic inheritance. The critical point is that the edge annotation scheme used in the gARG encoding allows us to associate..."

jeromekelleher: Good call, I'll update. I'll leave a pointer to the "Interpretation of Nodes" appendix also?

jeromekelleher:
I had another go at this, and tried to clarify.

BTW, I think we should try to move away from the specific phrase "edge annotations" as it's pretty opaque. Hopefully we can just say gARGs most of the time, and where we need to be precise talk about the "intervals of genetic inheritance associated with edges" or some such.

I'm going to merge this much as I need to go do other stuff now for a bit.

jeromekelleher:
Reading again, the last sentence needs another tweak, maybe end with "as in the eARG encoding" or something.

PR #264, opened 2022-09-29
Correct the caption for Fig3

hyanwong:
And references to subfigures


hyanwong:
This is obviously correct, to match the text up with #95, so I'm going to merge.

PR #265, opened 2022-09-29
Remove "local" as an edge annotation description

hyanwong:
We agreed at some point that it could be confusing to use "local inheritance" as a term, as it could get mixed up with "local trees". I prefer "direct-inheritance" (versus "samnple-resolved" inheritance). This PR attempts to reword it in that way.


jeromekelleher (approved):
LGTM

PR #266, opened 2022-09-29
Rejig and condense text about "big" ARGs etc

hyanwong:
I have tried to make the link between sample-resolving big ARGs, and (more importantly) forwards simulations, more explicit, as well as giving a tee-up into the simplification section.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -736,6 +736,9 @@ node (Fig~\ref{fig-ancestry-resolution}B). The second is to use the edge annotat
 describe the ``ancestral material''~\citep{wiuf1999ancestry,wiuf1999recombination}
 of the sample that is propagated through
 the graph from descendant to ancestor (Fig~\ref{fig-ancestry-resolution}C).
+As we discuss in section XXX, this second scheme has many advantages
+in terms of efficiency.
+
 
 \begin{figure}
 \centering
@@ -744,8 +747,8 @@ the graph from descendant to ancestor (Fig~\ref{fig-ancestry-resolution}C).
 The \citet[][Fig.~1]{wiuf1999recombination} example graph as an eARG and
 two gARGs with alternative edge annotation schemes.
 (A) A conventional Griffiths eARG. Square nodes represent events, with
-recombinations (in red) annotated with the position of a single breakpoint
-along the 7-unit long chromosome.
+each recombination (red) annotated with the position of a single breakpoint
+along the 7-unit-long chromosome.
 (B) A gARG with direct-inheritance edge annotations. Oval nodes represent
 genomes and annotations reflect ``local'' transmission of all genetic material
 between parent and child. Edges that do not span the full genome are visually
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -805,27 +809,45 @@ \section*{Inheritance and ancestral material}\label{Ancestry_resolution}
...
+ARGs such as that shown in Figs~{fig-ancestry-resolution}A~and~B
+is rarely simulated, because the resulting ``big ARG'' (Appendix XXX) is
+dominated by nodes such as \textsf{g} and \textsf{j} which contain information
+about inheritance in non-ancestral material, as well as nodes such as
+\textsf{o} and \textsf{q} which track details of ancestry even once the
+local genomic region has fully coalesced. Instead, simulating
+ARGs from the coalescent with recombination is usually done by
+running Hudson's algorithm to produce a sample-resolved ARG directly.
+Moreover, when inferring ARGs from genetic data, there is no information
+in the sample set that can be used to resolve these nodes, so that they
```
jeromekelleher, line 830: I would make this a bit stronger - they are entirely unknowable, because there is no information in the sample about them by definition.

paper.tex
```diff
@@ -805,27 +809,45 @@ \section*{Inheritance and ancestral material}\label{Ancestry_resolution}
...
 (i.e.~annotating the edges in the simulated pedigree with the genomic
 intervals inherited by a child from a parent) and periodically ``simplifying''
 this structure to retain only ancestral relationships that are relevant
 to the ancestry of the current population. In many situations, using this approach to
 record the full genetic ancestry of the current population is orders of magnitude faster
-than the alternative of tracking
-sequences~\citep{kelleher2018efficient,haller2018tree}.
+than the alternative of tracking neutral genetic variation directly~\citep{kelleher2018efficient,haller2018tree}.
+
+It is worth noting that nodes \textsf{g},  \textsf{j}, \textsf{o} and \textsf{q}
```
jeromekelleher, line 845: We should probably not ask the reader to look at the figures so much, or expect them to keep them in their head. I've been trying to bring references to specific things in figures down to a bare minimum, with concrete discussions kept to the captions.

I think we can say this without referring to specific nodes, right?

jeromekelleher (approved):
LGTM, a few minor comments.

paper.tex
hyanwong: I think that's true, at least for g and f. But I'm not sure that e.g. the grand MRCA is entirely unknowable (I think you can prove it exists in the Griffiths model, even if you can't date it). These are simply unknowable from data on the samples. I guess there are other ways to know about non-ancestral nodes too. But I'm happy to rephrase - you think I should take out the "essentially"?

paper.tex
hyanwong: Yeah, good point. We're quite a long way from the figure now. I'll rephrase.

paper.tex
hyanwong: I'm going to leave this as it is, and we can pick up in a further pass if necessary.

PR #268, opened 2022-09-29
Change colour scheme

hyanwong:
If we have RE and red and CA as blue, that's more colourblind friendly. We can then use green for genomes (alliterative!) which will help in the cellular diagram too.

We can then change the colour scheme for coalescent ARGs so that nodes with lots of parents (representing lots of recombination) get redder, whereas nodes with only one parent are still green. I think this new scheme ticks all the boxes, although we could revert to black oval outlines in fig 3 if we don't like the proliferation of colours.


jeromekelleher:
Sure, let's merge to try it out.

PR #271, opened 2022-10-04
Add a caption to figs 4 and 5

hyanwong:
Involves reversing the coordinates in Fig 4. I tried squeezing the captions in without changing the size of the plots, but it might look a bit cramped.


jeromekelleher:
Happy to merge and see how we like it.

PR #314, opened 2023-09-15
Fixup intro

jeromekelleher:


hyanwong:
LGTM. Feedback from Nate was that we should slip in somewhere in the first 2 paras that an ARG can be thought of / projected as a series of local trees. The idea was to catch the interest of casual readers, who might be familiar with a tree, but less so with an ARG. The sentence about "a means of
 describing the sequence of correlated local trees that
 are the \emph{outcome} of some such ancestral process" was Ana and my attempt to do that. I do think this would be a useful addition somewhere in the introduction, simply to easy people in, but I'm not sure where. Perhaps you can think of a better place @jeromekelleher? I'm happy with any suggested phrasing / placement, TBH.

jeromekelleher:
My takes is it's too hard to get across the subtleties of trees vs ARGs without saying something actively wrong, so thought best to avoid. We might drop something in later though

hyanwong:
I take Nate's point that we want to related this to something more familiar to a population geneticist, through. We don't actually introduce the concept of local trees anywhere (I think)? We just assume it is known by the reader. A simple sentence to acknowledge this would be helpful.

Issue #337, opened 2023-10-06
Add edge annotations to simplification figure?


jeromekelleher:
From Gideon & Alex:

> Fig 4: Would it be possible to provide an annotated version of the ARG in panel A? Either in this figure or maybe in the supplement? I think it might be nice for readers to be able to scrutinize how the trees exactly knit together in the ARG and having these annotations would make that easier.

I guess adding a supplemental figure in which we annotate the edges of each of the ARGs would be possible? I think it would be too noisy to include in the fig.

What do you think @hyanwong?

hyanwong:
A supplementary annotated edge figure would work well, I think.

hyanwong:
Like this? 
![simplification_with_edges](https://github.com/tskit-dev/what-is-an-arg-paper/assets/4699014/89762062-57ae-4d1a-bb29-4c716e201742)

hyanwong:
(It doesn't have the local tree in, obviously)

jeromekelleher:
Nice!

hyanwong:
Leaving this open until we actually write a supplementary

jeromekelleher:
I don't think we'll have a separate supplement until much later in the process, certainly not before review. Let's just put it in as a figure at the end.

PR #362, opened 2023-10-11
Tidy Fig 6

hyanwong:
Put nodes in the same order in all figures, remove whitespace.


illustrations.py
```diff
@@ -844,5 +854,19 @@ def get_edge_colors(ts):
 cli.add_command(simplification)
 cli.add_command(inference)
 
+
+### JSON data below from running tskit_arg_visualizer on the tree sequences above
+
+tskit_arg_visualizer_json = dict(
```
jeromekelleher, line 860: This is pretty nasty - maybe we could save these as individual (formatted) JSON files? Massive single line JSON dumps are not very nice for git.

jeromekelleher (commented):
Looks great - lets save the JSONs out of the Python code though. Piping through jq should give reasonable formatting.

illustrations.py
hyanwong: Yeah - happy to save them as (separate) plain text files. I'm not keen to format the JSON though. I just want to copy and paste from within the `tskit_arg_visualiser`. If we change them, chance are the entire JSON will change anyway, so I don't think we should worry about git diffs (and they might seem massive visually, but they are not many bytes to store)

hyanwong: I think of it as git's fault if it doesn't like strings with no line breaks 😄

hyanwong: OK, now as separate json files.

hyanwong:
I was in the middle of changing the `simplification` plot to address #337, so there's also a small change rolled in to do with that (I didn't want to go back and undo it all).

PR #366, opened 2023-10-12
A pass at unary nodes

jeromekelleher:


hyanwong:
Thanks Jerome. This mostly looks good to me. Is it worth merging and following up any changes on overleaf?

paper.tex
```diff
@@ -913,49 +910,108 @@ \section*{ARGs and local trees}
 % set of SPRs that reconcile independently inferred
 % trees from different regions of the genome.
 
+\section*{Locally unary nodes}
+The arity of a node is the number of child nodes it has. There is an
```
hyanwong, line 914: Should we say "we define the arity ....". You could equally well define a parent arity ("in-degree") as a child arity ("out-degree"), right? So we are using a specific meaning here, which we have defined.

jeromekelleher: It's probably unambigious as we're talking about combined tree/graph arity, but I added a "We can define"

PR #376, opened 2023-10-13
Make the inference plots a bit less tall

hyanwong:
To fit main paper text below. The longer captions seems to have pushed the figure onto its own page


jeromekelleher:
That can be a bit unpredictable (depends on surrounding paragraphs too), so we might be able to make a bit taller again at some point. I'm a bit worried about Andrew's comment on the legibility of the second row, and whether making it more compact will make that worse.

hyanwong:
Good point. We could open an issue to track.

Issue #348, opened 2023-10-06
Add a tabular representation of a gARG somewhere


hyanwong:
In Fig 2 we have a tabular version of an eARG, but nowhere do we show a tabular representation of a gARG.

> t struck us as a weird to show the “proper” tabular encoding for an eARG but not for a gARG when we read it through in Sedburgh. Like we were showing how to do the wrong thing in more detail than we were showing how to do the right thing

Gregor also makes this point:

>since this paper advertises gARG as a data format, this actually becomes an important point or make a clear reference to data format specification!

gregorgorjanc:
I looked into this and came up with the two tables below. I would appreciate someone double checking I did not make a typo. It was pretty straightforward though, since I only really had to list the nodes `a-p` and then read of edges from figure 1B. Maybe that's why @jeromekelleher isn't fuss about writing down these tables!

Doing this, lots of questions came out! Of course;)

1) Where do we put these tables? Into `\section*{Supplementary figures}` and rename this section to `\section*{Supplementary tables and figures}` or just `\section*{Supplementary}`?

2) Do we show just the edge table or also the node table? In the definition of gARG, we say that we will not be using times.

3) If we do show the node table, do we list all the nodes in the pedigree in figure 1 or just those nodes that are accessible from sample genomes a-d? The figure 1A and 1B show all nodes.

```
\begin{tabular}{c|c}
Node & Time
\hline
$\noderef{a}$ & 0\\
$\noderef{b}$ & 0\\
$\noderef{c}$ & 0\\
$\noderef{d}$ & 0\\
$\noderef{e}$ & 1\\
$\noderef{f}$ & 1\\
$\noderef{g}$ & 1\\
$\noderef{i}$ & 1\\
$\noderef{j}$ & 2 - remove this row?\\
$\noderef{k}$ & 2\\
$\noderef{l}$ & 2 - remove this row?\\
$\noderef{m}$ & 3 - remove this row?\\
$\noderef{n}$ & 3\\
$\noderef{o}$ & 3 - remove this row?\\
$\noderef{p}$ & 3 - remove this row?\\
\end{tabular}

\begin{tabular}{c|c|c}
Child & Parent & Interval\\
\hline
$\noderef{a}$ & $\noderef{e}$ & $[0,2)$\\
$\noderef{a}$ & $\noderef{f}$ & $[2,10)$\\
$\noderef{b}$ & $\noderef{g}$ & $[0,10)$\\
[...]

PR #381, opened 2023-10-14
Adding tables for gARG #348 and other bits

gregorgorjanc:
I have added table for gARG to address #348.

I have also added labels for all sections. To the best of my ability, I replaced all `see Section XXX` or `see  Appendix XXX` with section labels. For this to make sense, I switched from `\section*{}` to `\section{}`. Otherwise there is no section number to refer to and LaTeX produces `see Section _blank_`.

I have also sorted Appendix Figure and Table style to `AX`, where `X` is a number. It's easy to change to `SX` but since we talk about appendices I used `AX`.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -83,7 +83,8 @@ and forms the basis of an efficient computational framework.
 \textbf{Keywords:} Ancestral recombination graphs
 
 \linenumbers
-\section*{Introduction}
+\section{Introduction}
+\label{sec-intro}
 Estimating the genetic genealogy of a set of genome sequences
 under the influence of recombination,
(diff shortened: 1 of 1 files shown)


jeromekelleher:
LGTM, thanks Gregor.

I'd been holding off on doing cross-refs because biologists seem to have some basis mistrust of numbered hierarchical documents, and journals insist on referring to sections by name. Using the automatic "nameref" (or whatever it is)  never works properly, as there's always some cludgy phrasing as a result, so I was just going to go through at the end and fix it up.

However, I much prefer this way of doing it, so let's just stick with this for the preprint and review, and deal with meeting  daft formatting requirements later!

PR #401, opened 2023-10-22
Place tool config in a single file

### LINKED ISSUE #399: Add (json?) file specifying inference params etc.
We currently say

> The version numbers of inference software, the specific inference parameters
> needed to reproduce the ARGs, and (where necessary) functions to convert outputs to tskit format,
> are available from our GitHub repository

I think it would help if we created a single (json?) file with the version numbers and parameters used - we could then simply read this file in the makefile for the tools and when running them in `make_example_outputs.py`

hyanwong:
And use more recent (Dec 2022) versions of software, updating figures where required.

Fixes #399


README.md
```diff
@@ -1,2 +1,39 @@
...
+```
+
+Then to recreate plots used in the paper, type `make clean && make`
+
+### Re-running inferences
+
+To also re-run the software used for the inference plots, the
+necessary inference tools need to be installed by running the makefile
+in the tools directory (note that this
+requires `python2` to be installed, to run ARGweaver conversion software)
```
jeromekelleher, line 21: Also to run ARGweaver itself, right?

jeromekelleher (commented):
LGTM - one minor question about what py2 is for

README.md
hyanwong: I think ARGweaver is a C++ binary, so arg-sample (the main routine) can be run without any python

hyanwong: Yeah - the inference command is

```
tools/argweaver/bin/arg-sample --sites examples/Kreitman_SNP.sites --output examples/argweaver_output/arg-sample --overwrite --smc-prime --popsize 1e6 --mutrate 5.49e-09 --recombrate 8.4e-09 --randseed 111 --iters 3 --sample-step 10000 --no-compress-output
```

Which doesn't require python2

jeromekelleher: Ah, I thought argweaver itself was py2. Cool, all good.

PR #403, opened 2023-10-23
Remove SI units

jeromekelleher:
units stuff wasn't really working, easier to just typeset by hand


hyanwong:
Thanks for doing that.

PR #406, opened 2023-10-23
Restructure "local trees" section

jeromekelleher:
A work in progress on the local trees section, trying to spell out why we think using a combinatorial encoding is useful and important.

Needs more work, but actually this seems quite an important insight...

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -872,21 +872,21 @@ desired.
 
 Any eARG can be converted to a sample-resolved gARG using the same machinery,
 and the two-step process is illustrated in Fig.~\ref{fig-ancestry-resolution}.
-The first step is to take the input eARG (Fig.~\ref{fig-ancestry-resolution}A)
-and convert all event nodes to genome nodes and add inheritance annotations to
-the gARG's edges (Fig.~\ref{fig-ancestry-resolution}B).
-If the node is a common ancestor node, we annotate the single
+The first step is to take the input eARG (Fig.~\ref{fig-ancestry-resolution}A),
+duplicate its graph topology, and then add inheritance annotations
+to each of the gARG's edges (Fig.~\ref{fig-ancestry-resolution}B).
+If a given node is a common ancestor event, we annotate the single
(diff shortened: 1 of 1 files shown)


paper.tex
```diff
@@ -986,60 +987,63 @@ \section{Ancestral material and sample resolution}
...
-To recover the local tree $\pi$ at position $x$, we begin by
-setting $\pi_u = -1$ for each $u\in N$. Then, for each sample
+% (We use ``local'' rather than
+% ``marginal''
+% % could list loads here, but not much point.
+% trees~\citep[e.g.][]{griffiths1996ancestral,minichiello2006mapping,kelleher2016efficient}
+% to avoid the potential confusion with the
+% statistical concept of marginal distributions.)
+A fundamental property of genetics is that a
+% TODO is this (genome) clarification helpful or confusing?
```
gregorgorjanc, line 997: I like it!

paper.tex
```diff
@@ -986,60 +987,63 @@ \section{Ancestral material and sample resolution}
...
+given DNA nucleotide is inherited from exactly one parent (genome),
+both at an organismal and cell-by-cell level
+(Appendix~\ref{sec-cell-lineages-and-args}).
+These chains of single-parent inheritance over time give rise,
+by definition, to a tree structure.
```
hyanwong, line 1002: I like this - it's really clear

hyanwong:
Personally, I think the idea of locally unary nodes is the main thing that makes people think that the local trees don't encode the entire ARG. When someone says "local trees" they think of each tree as only consisting of nodes that are branch points. If we lose the locally unary nodes, I can see that there isn't enough information to reconstruct the ARG (i.e. including recombination nodes).

jeromekelleher:
Merging an update to the local trees and the "unary" nodes section. It's a bit more formal now, but since there was lots of confusion, I think this can't be avoided.

Hopefully some basic questions about why we bother with formal definitions etc have been answered.

PR #415, opened 2023-10-26
Move Wiuf & Hein caveat

hyanwong:
I think it distracts the reader to put the caveat about "not a standard simulated ARG" at the end of the section. This is my attempt to shorten the text and put something up front when we introduce the example instead. I hope it makes Section 4 end on a stronger note.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -632,8 +632,14 @@ desired.
 \end{figure}
 
 Any eARG can be converted to a sample-resolved gARG using the same machinery,
-and the two-step process is illustrated in Fig.~\ref{fig-ancestry-resolution}.
-The first step is to take the input eARG (Fig.~\ref{fig-ancestry-resolution}A),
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I don't think this helps, and we should avoid using the term "Little ARG". I like the current text

hyanwong:
OK, although I note we do mention "Little ARG" higher up, which is why I thought we could reuse that term. The point that we are specifically picking this *in order* to show non-ancestral material doesn't quite come out in the current text, so I thought it would kill 2 birds with one stone.

I still think it could seem weak to end on the note that (a) we picked a weird/unrepresentative example, and then (b) but we did so for didactic reasons.  But I don't feel that strongly either way, which is why I did this as an easy-to-reject PR.

jeromekelleher:
I think it's vital that we make it clear we're not posing straw men

PR #420, opened 2023-10-27
Use correct recombination rate in Drosophila inference

hyanwong:
I realised that the recombination rate I was using was 3 times smaller than it should have been. I also took the opportunity to update Ne. This (of course) changes the sample taken from ARGweaver, and it also slightly changes the relate output, so a fair few things touching Fig 5 need tweaking.


hyanwong:
I will tweak the text after this is merged. Not much will change, fortunately.

PR #426, opened 2023-11-02
Compress Appendix D

jeromekelleher:
I compressed appendix D down to the two main points (I think) it was trying to make. Paragraph 1 is about tagging events, and paragraph two is about the reality of polytomies and mutational information required to distinguish them.

What I deleted (I believe) was redundant either because the information was already in the figure caption, or unecesseary/inaccurate argumentation.

Polytomies are not enabled by the gARG view - there is absolutely no reason you couldn't have them in an eARG. Statements about approximations have already been made elsewhere.

### DIFF
--- a/paper.tex
+++ b/paper.tex
@@ -70,7 +70,7 @@ the genealogical trees for
 % we're talking about in direct language understandable by population
 % geneticists.
 a sample of sequences vary along the genome.
-The structure capturing the details of these intricately interwoven paths of 
+The structure capturing the details of these intricately interwoven paths of
 inheritance is referred to as an ancestral recombination graph (ARG).
 New developments have made it possible to infer ARGs at scale,
 enabling many new applications in population and statistical genetics.
@@ -339,7 +339,7 @@ We will use these two sets of terms interchangeably.
 In many settings, nodes are dated, i.e.\ each
 node $u\in N$ is associated with a time $\tau_u$.
 A node can represent any genome along a
-% A gamete *is* one of the genomes along a chain of cell divisions? 
+% A gamete *is* one of the genomes along a chain of cell divisions?
 % I don't see what adding a gamete clarifies, and we cite Hudson plenty
 chain of cell divisions~\ref{sec-cell-lineages-and-args},
 or can be interpreted as representing one of the genomes of an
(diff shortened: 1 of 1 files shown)


jeromekelleher:
@hyanwong can I get your thoughts here?

paper.tex
```diff
@@ -339,7 +339,7 @@ \section{Genome ARGs}
 In many settings, nodes are dated, i.e.\ each
 node $u\in N$ is associated with a time $\tau_u$.
 A node can represent any genome along a
-% A gamete *is* one of the genomes along a chain of cell divisions? 
+% A gamete *is* one of the genomes along a chain of cell divisions?
 % I don't see what adding a gamete clarifies, and we cite Hudson plenty
 chain of cell divisions~\ref{sec-cell-lineages-and-args},
 or can be interpreted as representing one of the genomes of an
```
hyanwong, line 345: We say "can be interpreted as representing one of the genomes of an individual". A gamete isn't a cell in an individual (it's its own thing), which is why I thought we should add this.

paper.tex
```diff
@@ -2470,47 +2470,54 @@ \section{Cell lineages and ARGs}
...
+two node representation seen in
+Figs.~\ref{fig-simplification} and \ref{hudson_vs_bigARG},
+whereas tagging the genome
+below a recombination event leads to the more
+conventional graphs in Figs.~\ref{fig-ancestry-resolution} and \ref{fig-inferred-args}A,B.
+
+%% JK:  What does this add that is not in caption?
+% Thick black lines trace an ARG that describes the ancestry of the lower arm of
+% the four sampled chromosomes, and these lines pass through gARG nodes (green
+% chromosomal regions). A number of these nodes have only one parent and one
```
hyanwong, line 2493: Yes, you are right - I made a slight clarification to the caption just to make sure. I think the pass-through thing can be omitted.

paper.tex
```diff
@@ -2470,47 +2470,54 @@ \section{Cell lineages and ARGs}
...
+The schematic illustrates an important point about the biological reality of polytomies.
+Three linages coalesce in the left-hand genome of individual $D_{10}$, but do
+so as the result of two successive bifurcations. This is \emph{necessarily} so,
+because the only known method of reproducing DNA is by (semi-conservative) duplication.
+Whether this polytomy is resolvable depends on the available mutational data.
+Mutations can occur along any cell lineages. For example,
+a mutation in the first cell division of $D_{10}$ could be shared between
+the two gametes produced by the cells in the left half of $D_{10}$
+but not shared by the right hand gamete.
+With enough mutations, each
```
hyanwong, line 2506: Good phrasing: it's neater to say "enough mutations" rather that "high rate or long genome"

paper.tex
```diff
@@ -2470,47 +2470,54 @@ \section{Cell lineages and ARGs}
...
+
+% Note that mutations can occur along any lineage in Fig.~\ref{fig-cell-lines}. This means, for example,
+% that a mutation in the first cell division of $D_{10}$ could be shared between
+% the two gametes produced by the cells in the left half of $D_{10}$ but not shared by
+% the right hand gamete. With a high enough mutation rate (or equivalently, a long enough genome), each
+% round of mitotic germline genome duplication within a single multicellular organism could
+% potentially be distinguished.
+% This implies that in ARGs that have been inferred from real data, polytomies represent a lack
+% of mutational information (rather than marking a single ``multifurcation event'').
+% The ability to represent polytomies, and conceptualise them as a (deliberately) imprecise statement about the order
```
hyanwong, line 2518: I do quite like these last 3 lines, to bring the topic back into the purview of the gARG and the paper in general. It might not be obvious that the gARG is a natural fit to this "world of imprecision", or that this is what we claim a polytomy in a gARG is representing?

I guess one point here is that if we try to infer an eARG from real data, but allow polytomies, we are implicitly "doing the gARG thing" of showing the outcome of the process, rather than the process itself. I was trying to imply that, but I didn't say it exactly.

hyanwong (approved):

hyanwong:
> @hyanwong can I get your thoughts here?

Mostly looks good to me, thanks for shortening.

My only comment is the last sentence: I think it is useful to remind people here that the reason we are talking about this is because of gARGs. I think in the new version we don't mention gARGs at all, right? But the point is that a polytomy in an inferred ARG is implicitly representing a summary of multiple events, so polytomies fit naturally into the "deliberately imprecise" gARG framework we have been talking about in the rest of the paper. (True, you could argue that eARGs could be tweaked to represent simulated polytomies, but I think we don't even need to go there: we just say that the situation in individual D_10 is a natural fit for gARGs)

jeromekelleher:
We shouldn't be arguing for anything one way or another here, it's an informational appendix intended to illustrate fundamental points about biology. Putting in stuff about gARGs and eARGs just obfuscates this. We make the point about using polytomies to represent uncertainty in the Discussion and elsewhere.

hyanwong:
I still think it is worth mentioning gARGs here, but maybe we will have to agree to disagree.

I could read the appendix as it stands and not realise that it is is any way relevant to the gARG encoding. That seems wrong to me.

jeromekelleher:
> I could read the appendix as it stands and not realise that it is is any way relevant to the gARG encoding. That seems wrong to me. I'm not saying we make an argument here, just that we point out a link to gARGs.

None of the other appendices are about gARGs. They're not intended as part of the central narrative of the paper. We make the points that you discuss repeatedly, so let's leave it there.

jeromekelleher:
> Did you see my comment above about "gametes" in the main text? I don't think we do mention anywhere that Hudson (and others) define ARGs in terms of gametes. We probably should somewhere, right?

I'm looking at that on overleaf

hyanwong:
> > I could read the appendix as it stands and not realise that it is is any way relevant to the gARG encoding. That seems wrong to me. I'm not saying we make an argument here, just that we point out a link to gARGs.
> 
> None of the other appendices are about gARGs. They're not intended as part of the central narrative of the paper. We make the points that you discuss repeatedly, so let's leave it there.

My inclination, if there is disagreement, is to leave stuff out, so you are probably right here!

PR #474, opened 2024-06-26
Figlabels

hyanwong:
Update node labels to uppercase to avoid confusing with subfigure labels. Also remove reference to colour in captions.

Add the supplementary pdf too


jeromekelleher (approved):
Looks good.

Are you going to resubmit the latex and all the figures for processing or make all the changes again manually in the proofing software?

hyanwong:
Thanks for merging. I've done all the changes again manually in their awful proofing software, as they had already made changes e.g. to American spellings and house grammar style. This PR doesn't include those, so the repo PDF isn't entirely in sync with the Genetics proof, but I don't think we want it to be anyway (otherwise we can't put it e.g. on bioRxiv if we wanted to, as it includes the proofreaders work too)

jeromekelleher:
I don't think we can put it on biorxiv now, it's only the accepted version.

jeromekelleher:
I don't think it's worth risking copyright infringement for a few minor aesthetic tweaks.

hyanwong:
Yeah, I agree. No reason to put it up there if the article is out anyway.

